PR #5, opened 2021-10-01
SMILES as REMARK in PDBQT to export RDKit mol

diogomart:
- PDBQTMolecule.export_rdkit_mol() exports an RDKit molecule
   without needing to use an SDF/MOL2 file as template, by using
   a SMILES string that is included as a REMARK in the PDBQT
 - now depend on RDKit
 - PDBQTMolecule __init__ is from string, not PDBQT filename
 - PDBQTMolecule.from_file() to init from filename


althea-hansel (approved):

diogomart:
Thank you @althea-hansel

PR #20, opened 2022-06-02
bug fix for arg smiles template

althea-hansel:
Original arginine smiles template was for neutral sidechain. Changed to represent charged sidechain.

Also catch some RDKit exceptions resulting from errors in assigning bonds from template, raise RuntimeError with more information.


diogomart:
Closing because we will abandon hard-coded SMILES.

PR #48, opened 2023-05-23
Reactive docking

diogomart:


diogomart:
implement #58 before merging

PR #90, opened 2024-03-05
reject rdkit mols with queries

diogomart:


diogomart:
Superseeded by PR #120

PR #91, opened 2024-03-22
Gracefully handle `nan` or `inf` charges in `molsetup.py`

amorehead:
* Gracefully handles `nan` or `inf` charges that may arise in `molsetup.py` due to using unsupported RDKit atom types

### DIFF
--- a/meeko/molsetup.py
+++ b/meeko/molsetup.py
@@ -928,6 +928,11 @@ class RDKitMoleculeSetup(MoleculeSetup):
                 if atom.GetAtomicNum() == 34:
                     atom.SetAtomicNum(16)
             rdPartialCharges.ComputeGasteigerCharges(copy_mol)
+            # zero-out any `nan` or `inf` charges that may arise from unsupported RDKit atom types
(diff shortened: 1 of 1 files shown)


diogomart:
The problem with this is that it fails silently, problematic molecules will go unnoticed, Currently, the PDBQT writer raises the error if there are NaNs or Infs, and the error can be escaped if `bad_charges_ok=True` which is useful for docking with Vina because charges are not needed.

amorehead:
I see. I did not notice `bad_charges_ok` the first time around, however, it seems like this will solve my initial issue. In which case, I'll close this PR. Thanks for your quick response!

diogomart:
Thank you for looking into it. Maybe `bad_charges_ok` should become `set_bad_charges_to_zero` so we don't see `nan` or `inf` in the PDBQT.

amorehead:
I think that might make more sense, since to me it seems like having `nan`s or `inf`s listed in the PDBQTs might also confuse users. In fact, I've made a PR for this here: https://github.com/forlilab/Meeko/pull/92.

PR #92, opened 2024-03-23
Change `bad_charge_ok` to `set_bad_charges_to_zero` in `writer.py`

amorehead:
* Changes `bad_charge_ok` to `set_bad_charges_to_zero` in `writer.py` to reduce potential user confusion caused by exporting PDBQT files with (potentially) `nan` or `inf` charge values.

### DIFF
--- a/meeko/writer.py
+++ b/meeko/writer.py
@@ -317,7 +317,7 @@ class PDBQTWriterLegacy():
         return atom_name, res_name, res_num, chain
 
     @classmethod
-    def _make_pdbqt_line_from_molsetup(cls, setup, atom_idx, count):
+    def _make_pdbqt_line_from_molsetup(cls, setup, atom_idx, count, set_bad_charges_to_zero=False):
         """ """
         pdbinfo = setup.pdbinfo[atom_idx]
         if pdbinfo is None:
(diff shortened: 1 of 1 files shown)


diogomart:
@rwxayheee I saw https://github.com/rwxayheee/Meeko/commit/e222e4014276f38450fdca00929da1aff9f8ae3b

What do you think of this PR?

My concern against converting nan or inf to zero is that zero is a valid partial charge. A pdbqt or molsetup filled with zeros can be interpreted as a successful assignment of partial charges.

Assigning zero on a per atom basis is not a good idea because then, if any of the atoms were to have a non-zero charge, it would be impossible to know something went wrong in the charge assigment (I don't think this happens though).

In any case, if charge assignment fails, I think we should raise an error or keep the nans, but not "pretend" charge assignment worked OK by setting charges to zero.

diogomart:
The electrostatic surface potential results from the partial charge of many atoms because electrostatics are long range. If the charge assignment fails for some atoms of a system, then it should be considered that it failed for the entire system. In light of this, in a scenario where the charge assignment resulted in NaN or Infinity for some atoms, setting a partial charge of zero on a per-atom basis would be misleading as it could be interpreted as a correct charge assignment. If the user does not care about charges at all, then the charge model can be set to "zero" (I'll expose this option to mk_prepare_ligand in a separate PR as it is only available from Python at the moment). For this reason I'm closing this PR.

diogomart:
#220 adds option `--charge_model` to mk_prepare_ligand where model `zero` can be specified

PR #106, opened 2024-04-26
optionally allow breaking macrocycle bonds with A atom types

diogomart:
Some macrocycles don't have a bond between two atoms of type C, but often there's a C-A bond. For example: https://www.rcsb.org/ligand/5P8 This PR allows breaking bonds between atoms of type A, including A-A bonds, with the caveat that parameters of C will be used for A atoms of broken bonds.

### DIFF
--- a/meeko/macrocycle.py
+++ b/meeko/macrocycle.py
@@ -11,7 +11,7 @@ from operator import itemgetter
 
 
 class FlexMacrocycle:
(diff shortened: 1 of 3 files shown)


psharma-sr (approved):
Straightforward, what I said about testing applies, assuming it works as intended it should be fine.

diogomart:
Thanks!

PR #107, opened 2024-04-26
API CHANGE PDBQTReceptor init from string

diogomart:
Example need: parse the receptor blob stored in a ringtail database.


psharma-sr (approved):
Looks good, pretty straightforward! Assuming you've done whatever testing you think is necessary to check it works, Approved!

diogomart:
Thanks!

PR #109, opened 2024-04-30
Refactoring Molecule Setups

psharma-sr:
A complete restructure of the `MoleculeSetup` class (including subclasses like `RDKitMoleculeSetup` and `OBMoleculeSetup`) to make MoleculeSetups more readable, more maintainable, and clean up inheritance between MoleculeSetups and subclasses.

- Adds several helper dataclasses to group and manage data in the `MoleculeSetup` class and eliminate synchronized OrderedDicts. These include the `Atom`, `Ring`, and `Bond` dataclasses.
- Enforces stricter typing on most variables, functions, and methods.
- Enforces default values for all classes
- More cleanly separates `MoleculeSetup` and `RDKitMoleculeSetup`, makes sure to isolate all methods and variables that depend on RDKit to be assigned or to work to `RDKitMoleculeSetup`.
- More clearly establishes external toolkit subclass extension requirements
- Cleans up JSON serialization and deserialization for `MoleculeSetup` objects and subclasses,

What remains to be done:

- Examine attributes like `flexibility_model` and `atom_params` to see if they can be converted from dictionaries to something more easily managed and checked.
- Clear out lesser used functionality
- Finalize documentation by double checking TODOs with team.

### DIFF
--- a/meeko/molsetup.py
+++ b/meeko/molsetup.py
@@ -18,6 +18,7 @@ from rdkit.Chem import rdPartialCharges
 from .utils import rdkitutils
 from .utils import utils
 from .utils.geomutils import calcDihedral
+from .utils.pdbutils import PDBAtomInfo
 from .receptor_pdbqt import PDBQTReceptor
 
 try:
@@ -65,6 +66,9 @@ class MoleculeSetup:
 
     Attributes
     ----------
+    mol: rdkit.Chem.rdchem.Mol
+        not guaranteed to be present in the object, contains the RDKit mol object that the molecule setup was based on
+        or corresponds to.
     atom_pseudo: List[]
         a list of indices of pseudo-atoms?
 
@@ -98,7 +102,7 @@ class MoleculeSetup:
         a mapping from atom index (int) to a list of neighboring atom indices?
     bond: OrderedDict()
         a mapping from bond id (int?) to a dictionary containing two elements. The elements are
-        bond_order (int) and rotable (bool).
+        bond_order (int) and rotatable (bool).
     element: OrderedDict()
         represents the atomic number of the atom index. A mapping from atom index to atomic number
     interaction_vector: OrderedDict()
@@ -673,7 +677,7 @@ class MoleculeSetup:
     # interaction vectors
     def add_interaction_vector(self, idx, vector_list):
         """ add vector list to list of directional interaction vectors for atom idx"""
-        if not idx in self.interaction_vector:
+        if idx not in self.interaction_vector:
             self.interaction_vector[idx] = []
         for vec in vector_list:
             self.interaction_vector[idx].append(vec)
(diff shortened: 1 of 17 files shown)


meeko_temp_task_refinement/MOLSETUP_NOTES.md
```diff
@@ -0,0 +1,92 @@
...
+
+* Modify so there are no more dictionaries and only lists/arrays.
+  * methods like `add_x`, `delete_x`, `get_x` wil deal with indexing
+* Provide infrastructure to calculate "differential"
+modifications, like tautomers: the class calculating tautomers will copy and modify the setup 
+to swap protons and change the atom types (and possibly rotations?), then
+there will e a function to extract only differences (i.e., atom types and coordinates,
+in the case of tautomers) and store them to make it a switch-state
+* Only use 1-dimensional arrays (C)
+* Update `add_atom` to not specify neighbors, which should be defined only when a bond is created?
```
psharma-sr, line 18: What did this mean/ is this something I should cover as we're making these changes

meeko/molsetup.py
```diff
@@ -4,623 +4,451 @@
...
-                element = atomic_nrs[i], 
-                pdbinfo = pdbinfo,
-                chiral = False,
-                ignore = False,
-                add_atom_params = {
-                    "rmin_half": vdw_by_atype[atype]["rmin_half"],
-                    "epsilon": vdw_by_atype[atype]["epsilon"],
-                }
+    # region Manually Building A MoleculeSetup
+    def add_atom(
```
psharma-sr, line 300: I've implemented the Atom structure mostly so we are not doing a lot more overhead managing indices between attributes/making sure attributes don't get out of sync. With this, we can also define null values for attributes that aren't currently set, or we can leave it as-is. The extent we want the atom structure to be buried varies, I can completely bury it/allow external access to do things like loop over all the elements in index order, if that seems fair.

meeko/molsetup.py
```diff
@@ -4,623 +4,451 @@
...
-        return molsetup 
+        Parameters
+        ----------
+        pdbinfo
+        charge
+        atom_type
+        is_ignore
+        anchor_list
+        rotatable
+        directional_vectors
```
psharma-sr, line 405: I was having a hard time seeing where directional vectors were being used? Do we still need them, and if so, what should be the supported datatypes and actions for them?

meeko/molsetup.py
```diff
@@ -4,623 +4,451 @@
...
-            raise RuntimeError("{len(param_idxs)} parameters but {len(self.atom_type)} in molsetup")
-        for i, j in enumerate(param_idxs):
-            self.atom_type[i] = f"{prefix}{j}"
-        return None
+        Returns
+        -------
 
+        """
+        # Places the atom at the end of the atom list.
+        pseudoatom_index = len(self.atoms)
```
psharma-sr, line 420: Earlier, the atom index assigned to a pseudoatom was the true atom count + the number of pseudoatoms, because of how the dictionary structure for the attributes worked. I changed it to this because this seemed to be the intention, is there any reason to change it back/to preserve fields like true_atom_count now that we have 1. lists 2. the atom structure and is_pseudo flag

meeko/molsetup.py
```diff
@@ -4,623 +4,451 @@
...
-        self.rmsd_symmetry_indices = ()
+        # Tracking atoms and bonds
+        self.atoms: list[Atom] = []
+        self.bond_info = {}
         self.rings = {}
-        self.rings_aromatic = []
-        self.atom_to_ring_id = defaultdict(list)
-        self.ring_corners = {}  # used to store corner flexibility
-        self.name = None
         self.rotamers = []
```
psharma-sr: Is this where we really want rotamers? Does this look like an appropriate level, or should they be specific to RDKit or something like that

meeko/molsetup.py
```diff
@@ -4,623 +4,451 @@
...
-
-    def set_atom_type(self, idx, atom_type):
-        """ set the atom type for atom index
-        atom_type : int or str?
-        return: void
+        blank_atom = Atom(atom_index, is_dummy=True)
+        self.atoms[atom_index] = blank_atom
+        return
+
+    def add_bond(
```
psharma-sr, line 477: For Get_Bond, do we want it to raise an error if the specified bond does not exist, return None with no warnings or errors, or return none with a warning?

meeko/molsetup.py
```diff
@@ -4,623 +4,451 @@
...
+        ----------
+        atom_index: int
+        overwrite: bool
+        pdbinfo: str
+        charge: float
+        element: int
+        atom_type: str
+        is_ignore: bool
+        is_chiral: bool
+        graph: List[List[int]]
```
psharma-sr, line 332: There seem to be some implementations in the original molecule setup that use graph to refer to the overall attribute, but neighbors to refer to the connected nodes specific to an atom. Would it be more prudent to just call the attribute neighbors?

meeko/molsetup.py
```diff
@@ -4,623 +4,451 @@
...
+
+# TODO: RENAME THIS TO NOT BE WORD VOMIT, CLEAN UP - Sets all the requirements for if you're building
+# A moleculesetup with an external toolkit like RDKit or OB
+class MoleculeSetupExternalToolBuild(ABC):
 
-    def get_bonds_in_ring(self, ring):
-        """ input: 'ring' (list of atom indices)
-            returns list of bonds in ring, each bond is a pair of atom indices
+    @staticmethod
+    def are_fourier_series_identical(fs1, fs2):
```
psharma-sr: Do you think this belongs to the molecule setup external tool base class or to the RDKit MoleculeSetup?

meeko/molsetup.py
```diff
@@ -4,623 +4,451 @@
...
-
-            The atom is built using the `walk_recursive` method
-        self.ring_atom_to_ring_id: dict
-            mapping of each atom belonginig to the ring: atom_idx -> ring_id
+        NOT MODIFIED YET
         """
-
-        def isRingAromatic(ring_atom_indices):
-            for atom_idx1, atom_idx2 in self.get_bonds_in_ring(ring_atom_indices):
-                bond = self.mol.GetBondBetweenAtoms(atom_idx1, atom_idx2)
```
psharma-sr, line 604: This seems like it's an RDKit specific ring thing, so to what extent should I have rings implemented in MoleculeSetups vs in RDKitMoleculeSetups/the external tool interface?

meeko_temp_task_refinement/MOLSETUP_NOTES.md
diogomart: It has been fixed meanwhile, so nothing left to do anymore. Not sure when it was fixed. In v3.0.0 there was still a `neighbors` argument in `add_atom`: https://github.com/forlilab/Meeko/blob/26e9ba24bdccd3e8ad766201b7d6d9d3a1a916ec/meeko/setup.py#L145

meeko/molsetup.py
diogomart: Yep, we don't need them.

meeko/molsetup.py
diogomart: Having the `is_pseudo` flag suffices. The count of "true" atoms is then the number of atoms which are not pseudos. I also think we need only one of `is_pseudo` and `is_dummy` as these are synonyms.

meeko/molsetup.py
diogomart: I think this is the right place.
Rotamers define values (angles) that are associated with any number of bonds. For example, we can have a molecule with 3 rotatable bonds, and have any number of rotamers. Each rotamer can define angles for any number of bonds. For example:
rotamer_1: bond_1=1.0, bond_2=1.5
rotamer_2: bond_1=3.0

meeko/molsetup.py
diogomart: I think raise error

meeko/molsetup.py
diogomart: Not sure I fully understood, but I think the answer is yes. Each atom should know about it's immediate neighbors, not the neighbors of every other atom. I see the full graph at the level of the entire molecule, not at the level of each atom. Moreover, the graph is really just the bonds, it duplicates the same information that the bonds already have. I think we should remove the graph, and the methods that currently use the graph would instead use the bonds.

diogomart: The neighbors are also redundant with bonds, but maybe it's convenient to keep both.

meeko/molsetup.py
diogomart: I would put this in the base MoleculeSetup, it could, in theory, be used for a manually constructed (non-RDKit) molsetup.

meeko/molsetup.py
diogomart: Absolutely, we can get rid of rings. And of other stuff:

`rings_aromatic` should be removed because FlexMacrocycle only needs to know whether or not each bond is rotatable. Aromatic bonds can't be broken anyway because they are set as non-rotatable. In the hypothetical scenario in which an aromatic bond is deliberately set as rotatable, then I'd argue that FlexMacrocycle should consider it as breakable.

`rings` should be removed because `FlexMacrocycle.collect_rings()` can invoke enumeration of the rings, no need to store the rings in an attribute of molsetup.

`is_chiral` was only needed to discourage FlexMacrocycle from breaking bonds in which one or both atoms are chiral, but this is only needed for the original method (https://pubs.acs.org/doi/full/10.1021/ci700036j) that did not use pseudo-atoms to preserve the geometry of the bond. With the current method, it is OK to break bonds involving chiral atoms.

`bond_order` is only ever used as a proxy for `is_rotatable` in convoluted ways. We are better off setting `is_rotatable` directly. The one exception is FlexMacrocycle being able to break double bonds, but this is questionable because we can't properly restore the bond geometry without a dihedral potential, so we probably better off breaking bonds that are rotatable

diogomart:
There were two commits left hanging (55524a1 and 4ee45b4), which is why this PR wasn't merged automatically with #149, but I think all changes have been been incorporated. @psharma-sr is this correct?

PR #114, opened 2024-05-03
Macrocycle simplify

diogomart:
Enables us to get rid of `bond_order` and `is_chiral` attributes in molsetups.


diogomart:
oops the first two commits weren't intended, but wanted to add those sooner or later

psharma-sr (approved):

Issue #118, opened 2024-05-13
Error when converting molecules derived from multimolecular archives


davidoskky:
When providing more than one molecule, only one is converted to PDBQT but the smiles remark contains both molecules.
This could be considered incorrect input, however it does not raise an error when generating a PDBQT file but only when converting back to SDF.

I feel it would be better to immediately signal the input is not correct.

To Reproduce

```
cat "c1ccccc1.CCC(Cl)CCC" > test.smi
obabel test.smi -O test.sdf
mk_prepare_ligand.py test.sdf -o test.pdbqt
mk_export.py test.pdbqt -o output.sdf
```

```
Traceback (most recent call last):
  File "/bin/mk_export.py", line 93, in <module>
    output_string, failures = RDKitMolCreate.write_sd_string(
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.11/site-packages/meeko/rdkit_mol_create.py", line 360, in write_sd_string
    mol_list = RDKitMolCreate.from_pdbqt_mol(pdbqt_mol, only_cluster_leads)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.11/site-packages/meeko/rdkit_mol_create.py", line 196, in from_pdbqt_mol
    mol = cls.add_pose_to_mol(mol, coordinates, index_map) 
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.11/site-packages/meeko/rdkit_mol_create.py", line 294, in add_pose_to_mol
    raise RuntimeError("Only H allowed to be in SMILES but not in PDBQT")
RuntimeError: Only H allowed to be in SMILES but not in PDBQT

```

rwxayheee:
> only one is converted to PDBQT but the smiles remark contains both molecules

Can confirm this, for smilar Smiles, obabel might generate a [multi-fragment PDBQT file](https://github.com/openbabel/openbabel/issues/2665#issuecomment-1933749283) which is also not instantly ready for docking. 
[...]

Issue #62, opened 2023-08-08
mk_export.py Runtime error


eightmm:
Hi!
I tried to convert dlg, the result file of AutoDock-GPU, to sdf by using mk_export.py. At this time, the following error was output.

![스크린샷, 2023-08-08 14-34-49](https://github.com/forlilab/Meeko/assets/78335868/5532889e-33b1-4a3f-b52a-6d052cece055)

Looking at DLG and PDBQT, I found that there are 4 hydrogen atoms.
![스크린샷, 2023-08-08 14-36-07](https://github.com/forlilab/Meeko/assets/78335868/03d33294-a720-4cd8-baaa-82272d7827b9)

Is this problem caused by rdkit not recognizing hydrogen properly? or is there another reason?

Thanks!

diogomart:
Hi!
The explicit H in the REMARK SMILES line are unusual, my first guess is that the error is relates to that. Could you post the input ligand pdbqt as text so i can take a look tomorrow?

eightmm:
@diogomart 

![스크린샷, 2023-08-08 15-27-06](https://github.com/forlilab/Meeko/assets/78335868/af204867-5066-47a7-aa94-a73681022c85)

REMARK SMILES [H]c1c(N([H])[H])c([H])c2nc3c([H])c(N([H])[H])c([H])c([H])c3c([H])c2c1[H]
REMARK SMILES IDX 26 1 2 2 3 3 7 4 12 5 14 6 18 7 20 8 23 9 10 10 9 11 22 12
REMARK SMILES IDX 25 13 11 14 4 15 5 16 6 17 15 18 16 19 17 20
REMARK H PARENT
REMARK Flexibility Score: inf
ROOT
ATOM      1  C   UNL     1       8.982  23.182  49.516  1.00  0.00     0.012 A
ATOM      2  C   UNL     1       9.670  23.649  48.475  1.00  0.00     0.026 A
ATOM      3  C   UNL     1      10.665  22.851  47.794  1.00  0.00     0.034 A
ATOM      4  C   UNL     1      10.902  21.573  48.264  1.00  0.00     0.054 A
ATOM      5  C   UNL     1       9.939  17.864  51.143  1.00  0.00     0.054 A
ATOM      6  C   UNL     1       9.231  17.253  52.163  1.00  0.00     0.034 A
ATOM      7  C   UNL     1       8.222  18.030  52.852  1.00  0.00     0.026 A
[...]

PR #120, opened 2024-05-19
check molecule consists of a single fragment and has no queries

diogomart:
Fixes https://github.com/forlilab/Meeko/issues/118


diogomart:
Added commit to raise error for rdkit mols with queries.
Earlier, a fix for a specific type of query reported in #62 was implemented in 309183914225f501a06ae354fe153ae806839f23
Other types of queries have been observed in SD files (aka SDF) from PDBBind.
This added commit (309183914225f501a06ae354fe153ae806839f23) is a general "solution".

meeko/molsetup.py
```diff
@@ -1615,6 +1615,10 @@ def from_mol(
         if mol.GetNumConformers() > 1 and conformer_id == -1:
             msg = "RDKit molecule has multiple conformers. Considering only the first one."
             print(msg, file=sys.stderr)
+        if len(Chem.GetMolFrags(mol)) != 1:
```
nbruciaferri, line 1618: Looks good to me!

nbruciaferri (approved):

diogomart:
Thanks @nbruciaferri!

Issue #96, opened 2024-04-04
RDKit v2024 breaks meeko


Tonylac77:
After updating to rdkit v2024.03.1 there is a failed import (rdkit.six)

Traceback (most recent call last):
  File "/home/dockm8/miniconda3/envs/dockm8/lib/python3.10/site-packages/streamlit/runtime/scriptrunner/script_runner.py", line 535, in _run_script
    exec(code, module.__dict__)
  File "/home/dockm8/DockM8/gui.py", line 14, in <module>
    from scripts.docking_functions import DOCKING_PROGRAMS
  File "/home/dockm8/DockM8/scripts/docking_functions.py", line 12, in <module>
    from meeko import PDBQTMolecule, RDKitMolCreate
  File "/home/dockm8/miniconda3/envs/dockm8/lib/python3.10/site-packages/meeko/__init__.py", line 32, in <module>
    from .rdkit_mol_create import RDKitMolCreate
  File "/home/dockm8/miniconda3/envs/dockm8/lib/python3.10/site-packages/meeko/rdkit_mol_create.py", line 11, in <module>
    from rdkit.six import StringIO
ModuleNotFoundError: No module named 'rdkit.six'

diogomart:
Thank you for reporting. It is fixed in ed395cc8c4c690d25a295c1ba81ee0f38c620dd4 which can be used when installing from source (not in a release yet) https://github.com/forlilab/Meeko?tab=readme-ov-file#installation-from-source

PR #121, opened 2024-05-19
Fix import error by replacing rdkit.six with io.StringIO

SergeiNikolenko:
This pull request addresses an import error caused by the removal of `six` module in RDKit versions >= 2020.03.3. The import statement `from rdkit.six import StringIO` is replaced with `from io import StringIO` in `meeko/rdkit_mol_create.py`.

This change ensures compatibility with the latest RDKit versions and resolves the `ModuleNotFoundError` for `rdkit.six`.


diogomart:
Thanks! But it's fixed already, just not released.
ed395cc8c4c690d25a295c1ba81ee0f38c620dd4
#96

PR #123, opened 2024-05-22
Added option to put pdbqts into subdirectories

allisonbarkdull:


diogomart:
Modified to write zipped (.tar.gz) files with 10k PDBQTs each. Command line options are
- `-z` / `--multimol_targz` triggers tar-gzeeing
- `--multimol_targz_size` to set the number of PDBQTs per .tar.gz (default is 10k)

Additionally added `--name_from_prop` for SD files that store the molecule name in a property, such as `>  <mol_id>`, instead of storing in the designated name field, which is the first line of the MOL block.

diogomart (approved):

PR #133, opened 2024-06-13
Refactored receptor class (chorizo) to generalize beyond proteins

diogomart:
The changes are extensive, and include:
 - every chorizo residue has its own MoleculeSetup instance, whether flexible (sidechain) or not
 - refactored flexibility.py


meeko/__init__.py
```diff
@@ -47,21 +55,22 @@
...
-        'oids_block_from_setup',
-        'parse_offxml',
-        'Hydrate',
-        'Restraint',
-]
+__all__ = ['MoleculePreparation', 'RDKitMoleculeSetup', 'MoleculeSetupEncoder',
+           'pdbutils', 'geomutils', 'rdkitutils', 'utils',
+           'AtomTyper', 'PDBQTMolecule', 'PDBQTReceptor', 'analysis',
+           'LinkedRDKitChorizo', 'ChorizoResidue', 'ResiduePadder', 'ResidueTemplate',
+           'LinkedRDKitChorizoEncoder', 'ChorizoResidueEncoder', 'ResiduePadderEncoder', 'ResidueTemplateEncoder',
```
psharma-sr, line 63: ResidueChemTemplates and the ResidueChemTemplatesEncoder should probably be added here

meeko/data/residue_chem_templates.json
```diff
@@ -0,0 +1,567 @@
+{
```
psharma-sr, line 1: We should remember to include a description of what this data file is in documentation. I'm putting this comment here as a reminder, I'll add that to the documentation issue more explicitly and resolve this comment later.

meeko/flexibility.py
```diff
@@ -5,275 +5,269 @@
...
+        flex_model["rigid_body_connectivity"].pop((group_index, host_index))
+        flex_model["rigid_index_by_atom"][atom_index] = host_index
+    return
+
+
+def get_flexibility_model(
+    molsetup,
+    root_atom_index=None,
+    break_combo_data=None,
+    glue_pseudo_atoms=None,
```
psharma-sr: glue_pseudo_atoms is not used anywhere in this function, should it be deleted or should we keep it for future functionality

meeko/flexibility.py
```diff
@@ -5,275 +5,269 @@
...
-        return self.setup
-
-
-    def select_best_model(self):
-        """
-        select flexibility model with best complexity score
-        """
-        if len(self.flexibility_models) == 1: # no macrocyle open rings
-            best_model = list(self.flexibility_models.values())[0]
+def _calc_max_weighted_depth(model, seed_node, bonds_to_break, visited=None, depth=0):
```
psharma-sr: This file probably needs a few function level comments, but I think I can handle some of that to get it started/make it easier

meeko/flexibility.py
```diff
@@ -5,275 +5,269 @@
...
+    remove = {}
+    for group_index, members in members_dict.items():
+        if (
+            len(members) == 1
+            and len(graph[group_index]) == 1
+            and members[0] not in not_terminal_atoms
+            and group_index != flex_model["root"]
+        ):
+            remove[group_index] = members[0]
+    flex_model["rigid_body_count"] -= len(remove)
```
psharma-sr: Would we want to consider updating the count after the removal has happened? How likely is it that there would be an error during the removal for loop?

meeko/flexibility.py
```diff
@@ -5,275 +5,269 @@
...
+        )
+        if len(flex_model["visited"]) != nr_not_ignored:
+            molsetup.show()
+            for key, value in flex_model.items():
+                print(key, value)
+            msg = f"{len(flex_model['visited'])=} differs from not-ignored atoms {nr_not_ignored}"
+            raise RuntimeError(msg)
+        root_body_index = get_root_body_index(flex_model, root_atom_index)
+        flex_model["root"] = root_body_index
+        broken_bonds = []
```
psharma-sr, line 168: Is there a reason we're bothering to return broken_bonds if it's an empty list always anyway?

meeko/flexibility.py
diogomart: Will remove, it's only needed in `update_closure_atoms` where it's called `coords_dict` (will rename to `glue_pseudo_atoms` for consistency). Just for reference here's a test for this feature: https://github.com/forlilab/Meeko/blob/6317966790b4e9817e37649bf771adb13bc3d921/test/macrocycle_test.py#L35

meeko/flexibility.py
diogomart: If there's an error mid loop and it's handled with an except statement, the entire molecule setup would be compromised and should be discarded. In any case, I'll update the count after the for loop in case someone ever thinks of using the count as a proxy for successful completion.

meeko/flexibility.py
diogomart: that's the first return statement of this function, the second one returns non-empty lists: https://github.com/forlilab/Meeko/blob/6317966790b4e9817e37649bf771adb13bc3d921/meeko/flexibility.py#L125-L126

meeko/flexibility.py
diogomart: added in ac2da0724891385abdebe0a207e7a1fb7a202453

meeko/flexibility.py
diogomart: added a couple in ac2da0724891385abdebe0a207e7a1fb7a202453

meeko/flexibility.py
diogomart: removed in ac2da0724891385abdebe0a207e7a1fb7a202453

meeko/flexibility.py
```diff
@@ -5,275 +5,279 @@
...
+        for ring in unbroken_rings_list[index]:
+            for bond in molsetup.get_bonds_in_ring(ring):
+                unbroken_rings_bonds.append(bond)
+
+        model = walk_rigid_body_graph(molsetup, bond_break_combo, unbroken_rings_bonds)
+        nr_not_ignored = sum(
+            [not molsetup.atom_ignore[i] for i in range(len(molsetup.atom_ignore))]
+        )
+        if len(model["visited"]) != nr_not_ignored:
+            msg = f"{len(model['visited'])=} differs from not-ignored atoms {nr_not_ignored}"
```
psharma-sr: Is this runtime error not being exactly like the first intended? Do we still need the print statement in the earlier RuntimeError?

meeko/flexibility.py
```diff
@@ -5,275 +5,279 @@
...
+    for index in range(len(bond_break_combos)):
+        bond_break_combo = bond_break_combos[index]
+        bond_break_score = bond_break_scores[index]
+        unbroken_rings_bonds = []
+        for ring in unbroken_rings_list[index]:
+            for bond in molsetup.get_bonds_in_ring(ring):
+                unbroken_rings_bonds.append(bond)
+
+        model = walk_rigid_body_graph(molsetup, bond_break_combo, unbroken_rings_bonds)
+        nr_not_ignored = sum(
```
psharma-sr, line 115: I don't think this would be changing with the loop, but maybe I'm missing something. Could it just be instantiated outside the loop?

meeko/flexibility.py
diogomart: That's right, the errors are identical. Removed print statements in 41247d2866a5b0a3bab4825685aeddbb1289d4a0

meeko/linked_rdkit_chorizo.py
```diff
@@ -226,291 +517,783 @@ class LinkedRDKitChorizo:
...
+            )
+            if len(missing):
+                raise ValueError(
+                    f"Residue IDs in set_template not found: {missing} {raw_input_mols.keys()}"
+                )
+
+        # currently allowing only one bond per residue pair
+        if any([len(v) > 1 for k, v in bonds.items()]):
+            msg = "got more than one bond between some residue pairs:" + os_linesep
+            for key, value in repeated.items():
```
psharma-sr: I don't think repeated is defined here, not sure what it should be though

meeko/macrocycle.py
```diff
@@ -231,15 +364,18 @@ def show_macrocycle_scores(self, setup):
             bond_by_ring = defaultdict(list)
```
psharma-sr, line 364: It doesn't look like this function is being used anywhere, so I have deprecated it for now. All this logic here is after a return statement and is therefore inaccessible code, so if we did want to keep this function we should give that some thought.

PR #153, opened 2024-08-20
Quick Fixes for Inconsistent Selection Formats in reactive_flexres and all_flexres

rwxayheee:
This PR contains three minor changes: 
(1) Quick Fixes for Inconsistent Selection Formats in reactive_flexres (dict where its keys are tuples) and all_flexres (list of strings); 
(2) Removed a Redundant Loop while Iterating args.flexres; 
(3) Minor Edits to a Help Message (--flexres) and an Error Message (Unsupported Ligand File Format)

@diogomart could you please confirm if "--reactive_flexres" is supposed to work (https://github.com/forlilab/Meeko/blob/11bc7d14f1c363b14289957fef542079ff39ea08/scripts/mk_prepare_receptor.py#L478-L479)? If it's not currently supported, then there's no rush to pull. I have no prior experience with reactive docking, so I'm not very sure what the expected outcomes look like.. 

This appears to be two commits because I accidentally included modifications in a locate fork (where I have a rewrite of `parse_cmdline_res` and `parse_cmdline_res_assign`). These changes were not necessary so I changed them back to their versions in the current develop branch.

### DIFF
--- a/scripts/mk_prepare_receptor.py
+++ b/scripts/mk_prepare_receptor.py
@@ -40,54 +40,49 @@ with open(templates_fn) as f:
 
 
 def parse_cmdline_res(string):
-    """ Convert "A:5,7,BB:12C"  to  ["A:5", "A:7", "BB:12C"] """
-    blocks = [s for s in string.split(",") if s]
-    chain = ""
-    res_set = set()
-
-    for b in blocks:
-        if ":" in b:
-            segs = b.split(":")
-            if len(segs) > 2:
-                raise ValueError(f"too many delimiters (:) in {b}")
-
-            chain_id, resnum = segs
-            if not chain_id:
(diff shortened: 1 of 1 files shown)


diogomart:
Thanks! Yes `--reactive_flexres` is supposed to work.

To make the reactive residues parsing consistent with the flexible residues we should use:
 - `parse_cmdline_res` for `--reactive_flexres`
 - `parse_cmdline_res_assign` for `--reactive_name_specific`
And then delete functions `parse_residue_string` and `parse_residue_string_and_name`

Also:
 - un-comment the `args.reactive_name` block on lines 400-411
 - no space for empty chain in help message, i.e. `":17"`, not `" :17"`

rwxayheee:
@diogomart Got it! I will close this then

PR #154, opened 2024-08-26
update function to write PDB from chorizo, and fix a bug that ignored input H positions 

diogomart:
Minor changes with regards to `LinkedRDKitChorizo.to_pdb()`, just keeping up to date with all the recent refactoring.

H position on link atoms, which are atoms that have bonds to adjacent residues, were being replaced by RDKit calculated positions. This is now fixed.


itaneja2 (approved):

diogomart:
Thanks :)

PR #162, opened 2024-09-09
rename mk_prepare_receptor.py arg --ligand to --box_enveloping

diogomart:
@rwxayheee does `--box_enveloping` make sense to you? Maybe there are better alternatives. `--ligand` was a bad choice I made back in the day.

### DIFF
--- a/scripts/mk_prepare_receptor.py
+++ b/scripts/mk_prepare_receptor.py
@@ -158,11 +158,11 @@ def get_args():
         action="store_true",
     )
     box_group.add_argument(
-        "--ligand",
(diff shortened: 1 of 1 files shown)


rwxayheee:
Hi @diogomart 
Yes, this makes sense to me. It implies that the file passed can be a ligand or a polymer (like a protein PDB file or part of the PDB file). This allows users to center gridbox at selected residues. `mol.GetConformer().GetPositions()` gets the coordinates of all atoms if mol came from `Chem.MolFromPDBFile`. I like the change, thank you!

rwxayheee (approved):

PR #165, opened 2024-09-14
mk_prepare_receptor output flags improvement

diogomart:


diogomart:
Removed `--skip_gpf`. Now, a .gpf file is not written by default, it is triggered by `-g/--write_gpf`.

Removed `--chorizo_config`.

Can run `mk_prepare_receptor` without writing any file, in which case a detailed message is printed to show the most common files to write and which flags to use.

Write flags can be used as a on/off flag without any argument, e.g. `--write_gpf`, or with a filename as argument: `--write_gpf rec.gpf`. The default base name is set with `-o/--output_basename`.

Now `-g` is now an abbreviation for `--write_gpf`, not for setting reactive names. 

Renamed:
`--pdb` to `--read_pdb`
`--macromol` to `-i/--read_with_prody`, but macromol is still used as metavar in help message
`--output_filename` to `--output_basename` and changed its behavior, it doesn't set the pdbqt output filename anymore, now `-p/--write_pdbqt` does that.

@rwxayheee let me know if you don't like any of this!

rwxayheee:
Thanks for the nice work. This is not a big issue - in the code block that generates the reactive docking parameters: 
https://github.com/forlilab/Meeko/blob/3c4e03053f48a56768bfa463b99ca1f52a911903/scripts/mk_prepare_receptor.py#L854-L927

`all_flex_pdbqt`, `rigid_fn` and `flex_fn` are not defined when PDBQT is not requested. This may result in NameError when the command line script is executed without '-p`, like with just`-o {output_basename} -j`. 

There are some different ways to fix it: 
(1) Make `-p` True by default (this will disable the json-only write option `-o -j`), or
(2) Give `all_flex_pdbqt`, `rigid_fn` and `flex_fn` some initial values (just to avoid NameError), or
(3) Deactivate the code block if PDBQT is not requested and/or GPF is not requested. In these cases, a reactive docking parameter file is probably not expected

I was doing (2), but I withdrew the change, thinking maybe (3) is better.,

rwxayheee (approved):
I like the design. Tested with combinations of write options for Vina and AD-GPU. I will approve it. @diogomart It would be great if you could take a quick look at json-only option (-o <name> -j), but it's not a very big issue

diogomart:
I agree (3) is best, implemented in df522677ac548b67ab72c8b3f2f34b57a9efa612. Glad you found this

PR #167, opened 2024-09-14
remove dropped attribute from molsetup.show()

diogomart:
The attribute `directional_vectors` has been removed from MoleculeSetup, but was still present in the `MoleculeSetup.show()` method.


althea-hansel (approved):

diogomart:
Thanks!

PR #172, opened 2024-09-22
Minor edit to atom_name_to_molsetup_index; Support input in both ways: --macromol (ProDy) and --pdb (RDKit)

### LINKED ISSUE #168: mk_prepare_receptor for reactive docking isn't working with --macromol (Prody parser)
The origin of this issue is

https://github.com/forlilab/Meeko/blob/5f29de0c7e8f0193c0673db1785b38196c24708b/meeko/reactive.py#L120

only works when RDKit's PDB parser is used. Currently, the easy walkaround is to use --pdb (RDKit parser). Happy to provide a simple fix (but can also wait until other major changes are made, if Prody is going to be optional in the near future!

rwxayheee:
This fixed #168 without further changes related to ProDy. 

Before (uses names from RDKit's GetPDBResidueInfo(), for --pdb only):
```py
def atom_name_to_molsetup_index(chorizo_residue, atom_name):
    indices = []
    for atom in chorizo_residue.raw_rdkit_mol.GetAtoms():
        name = atom.GetPDBResidueInfo().GetName().strip()
        if name == atom_name:
            indices.append(atom.GetIdx())
    if len(indices) > 1:
        raise RuntimeError(f"multiple atoms matched query atom name {atom_name}")
    if len(indices) == 0:
        return None
    index = indices[0]
    index = chorizo_residue.mapidx_from_raw[index]
```
 

After (uses names from parameterized chorizo residue): 
```py
def atom_name_to_molsetup_index(chorizo_residue, atom_name):

    # get matched indices from parameterized chorizo_residue
    indices = [index for index, name in enumerate(chorizo_residue.atom_names) if name == atom_name]
    
    if len(indices) > 1:
        raise RuntimeError(f"multiple atoms matched query atom name {atom_name}")
    if len(indices) == 0:
        return None
    
    index = indices[0]
```

### DIFF
--- a/meeko/reactive.py
+++ b/meeko/reactive.py
@@ -115,17 +115,16 @@ get_reactive_atype = reactive_typer.get_reactive_atype
 
 
 def atom_name_to_molsetup_index(chorizo_residue, atom_name):
-    indices = []
-    for atom in chorizo_residue.raw_rdkit_mol.GetAtoms():
-        name = atom.GetPDBResidueInfo().GetName().strip()
-        if name == atom_name:
-            indices.append(atom.GetIdx())
+
+    # get matched indices from parameterized chorizo_residue
+    indices = [index for index, name in enumerate(chorizo_residue.atom_names) if name == atom_name]
+    
     if len(indices) > 1:
         raise RuntimeError(f"multiple atoms matched query atom name {atom_name}")
     if len(indices) == 0:
         return None
+    
     index = indices[0]
-    index = chorizo_residue.mapidx_from_raw[index]
     inv = {j: i for i, j in chorizo_residue.molsetup_mapidx.items()}
     index = inv[index]
     return index


diogomart:
In this way, the atom name that will be used is the one in the templates. This has two problems:
1. it may confuse the user if the atom name in the input file differs from the one in the template
2. in the future, we might want templates without associated atom names

I think we should fix the prody parser instead.

rwxayheee:
Hi @diogomart 
Ok. Do you mean we should always use names associated with the "raw" (input) mols per residue? 
I can look at the details (prodyutils.py?) a little later today. From a quick glance I'm a bit confused why .GetPDBResidueInfo().GetName() isn't working with ProDy

diogomart:
> Do you mean we should always use names associated with the "raw" (input) mols per residue?

I think yes if the goal is to have the user specify an atom. Likely the user looked at the system in a molecular visualizer, and is using the atom name associated with the input file (pdb/mmcif).

PR #173, opened 2024-09-25
Small fix in prodyutils.py

### LINKED ISSUE #168: mk_prepare_receptor for reactive docking isn't working with --macromol (Prody parser)

rwxayheee:
The change is to assign rdkit_atom.SetPDBResidueInfo before mol.AddAtom(rdkit_atom). This seemed to fix #168 (and maybe the comment `# TODO passibg values at constructor not working?` in the code block?)

### DIFF
--- a/meeko/utils/prodyutils.py
+++ b/meeko/utils/prodyutils.py
@@ -419,8 +419,7 @@ def _prody_residue_to_rdkit(
         if keep_charges is True:
             partial_charge = atom.GetPartialCharge()
             rdkit_atom.SetProp("_prody_charges", partial_charge)
-        # add atom to the molecule
-        mol.AddAtom(rdkit_atom)
+        
         # store atom coordinates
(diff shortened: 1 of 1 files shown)


diogomart:
There are two functions that need the fix: `_prody_residue_to_rdkit` and `_prody_hierarchy_to_rdkit`. Most code is duplicated and we might not need to treat Prody residues and Prody selections differently because a for loop over either type iterates over atoms. @sforli is there a reason to treat selections differently from residues?

diogomart (approved):

PR #174, opened 2024-09-25
make prody optional

diogomart:


rwxayheee (approved):
I made a small change to mk_prepare_receptor because _prody_import_error was defined in except block, but it's required outside (further down the script). For mk_prepare_receptor.py it's ready to merge. 

For ligand preparation: 
I did a quick check looking for other attempts to import prody. In mk_prepare_ligand.py there is:
https://github.com/rwxayheee/Meeko/blob/03d4168a72cb0d52043852ae388f025e5461e84c/meeko/cli/mk_prepare_ligand.py#L18-L22

because **meeko/covalentbuilder.py** uses prody. This made me think that ligand preparation for covalent docking (tethered) requires prody. 

Do you also want to make prody optional for this? We can make another PR or leave this open until (someday) we add prody parser to covalentbuilder

diogomart:
It already is optional for tethered docking. Improved help messages here: 0619b7fe22d6a4ac7e43715fe20dab7770306a24

@rwxayheee please merge if it looks good to you.

PR #175, opened 2024-09-26
remove unneeded code for prody to rdkit conversion

diogomart:
All considered Prody objects (AtomGroup, Selection, Residue, Chain) yield Prody atoms when iterated over with a `for` loop. Hence, there is no need for separate functions to convert each of these types to an RDKit molecule. This commit simplifies the code by simply iterating over any of the prody types.

The types are different only with respect to the `getBonds()` method: only the `AtomGroup` implements it. Thus, the other types must call `getAtomGroup()` first. It should be noted that Prody ignores bonds from mmCIF files even if bonds are present.

### DIFF
--- a/meeko/utils/prodyutils.py
+++ b/meeko/utils/prodyutils.py
@@ -102,7 +102,6 @@ def rdkit_to_prody(
 
 
 ALLOWED_PRODY_TYPES = Union[Selection, AtomGroup, Chain, Residue]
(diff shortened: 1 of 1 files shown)


rwxayheee (approved):
I checked with some tests and it did not change the prepared files before vs. after. I approve the PR, thank you for looking in to this @diogomart

diogomart:
Thanks!

PR #176, opened 2024-09-27
Add nucleic acid residues to residue_chem_templates.json

rwxayheee:
This change is only about the chem templates (a data file). The existing nucleic acid residues data (ambiguous and residue_templates) were replaced. New data were generated from Chemical Component Dictionary (https://www.wwpdb.org/data/ccd) from a component cif file following [a gemmi -> rdkit -> json workflow](https://github.com/rwxayheee/gemile). 

The workflow might be useful to generate chemical templates for cofactors and other registered ligands for receptor and ligand preparation, when atom names are needed (to interface programs that need atom names for mapping like Amber). However, some more work is needed to enumerate protonation states, verify linker atoms, extend linking reactions, etc. 

Some example RNA receptors PDBQT with the modified data: 
[RNAReceptorPDBQT.zip](https://github.com/user-attachments/files/17157466/RNAReceptorPDBQT.zip)

They were prepared with `--allow_bad_res` to show normal template mapping vs. blunt end at linker atom vs. deletion. The examples include some special cases like (PDB 4LVZ) where X5 (5' end residue, typically a nucleoside - no phosphate) with an excess phosphate, which is termed X5p in the modified data. 

DNA receptors were not verified (I don't have examples on hand, but will do in the very near future!


rwxayheee:
Unfortunately, some more work is needed to check the inconsistent atom names and tautomeric forms in the source cif file. Currently, the following variants are considered: 

```py
# nucleotide
    variant_dict = {
        "_": {}, # free nucleotide monophosphate
        "":  {'OP3', 'HOP3', "HO3'"}, # embedded nucleotide 
        "3": {'OP3', 'HOP3'}, # 3' end nucleotide 
        "5p":  {"HO3'"}, # 5' end nucleotide (extra phosphate than canonical X5)
        }
```

and 

```py
# nucleoside (no phosphate)
    variant_dict = {
        "N": {}, # free nucleoside 
        "5":  {"HO3'"}, # 5' end nucleoside (canonical X5 in Amber)
        # there shouldn't be a 3' end nucleoside variant (always has a phosphate)
        }
```

The nucleotides are fine. But the nucleoside components have some inconsistent atom names. The free nucleoside of DG, GNG, is also a tautomer of the canonical DG. Therefore, it's better to make the needed nucleoside chemical components from the nucleotide templates. Variants with the same parent will be derived from one and only the nucleotide component. 

Because it's not safe to merge, I will close the PR for now. 

(actually, the new templates are ready as of [959268c](https://github.com/rwxayheee/Meeko/commit/959268cffe1b19b7e197f1644b95b0d22ac71e1e), as long as the atom names are not used... the standard names, canonical tautomers, naming for alternate protonation states, need to be discussed with different people who do NA simulations

rwxayheee:
[959268c](https://github.com/rwxayheee/Meeko/commit/959268cffe1b19b7e197f1644b95b0d22ac71e1e) was deprecated because chemical component [GNG](https://www.rcsb.org/ligand/GNG) was used to build the nucleoside of DG (2'-deoxy guanosine). In fact, it is a different tautomer of the canonical G, different from G/DG/GMP. In the following commit, [L1J](https://www.rcsb.org/ligand/L1J) was used instead. 

[52f0c0b](https://github.com/forlilab/Meeko/pull/176/commits/52f0c0b7c943af317f82745dacc06f86c7cbeaf4) was safe to merge, but the nucleosides might have inconsistent atom names and therefore not ideal. 

As of https://github.com/forlilab/Meeko/pull/176/commits/47652c0499dce1030ef928616f5e4d0262e3ccc3, all nucleosides were derived from nucleotides, a.k.a for individual nucleobases, there's one and only one parent chemical component to derive the variants with different embedding types. They are exactly - 'A', 'U', 'C', 'G', 'DA', 'DT', 'DC', 'DG'-  in CCD's components.cif. This is a working version. 

Further verification and addition of ion information are underway.

rwxayheee:
https://github.com/forlilab/Meeko/pull/176/commits/700396c338c7ae8432fe15cbc5e490cbf7ba26a9 has the following changes in addition to formatting: 

- Dropped free nucleotides/nucleosides (X_, XN) from ambiguous dict 

Most cases the template mapping was attempted on residues. It's just inefficient to match residues with 2 additional free molecular forms. Currently, we also don't have chem templates for free amino acids. 

- Added currently supported ions by AD4

The charge state of Fe is let to be +3 to be consistent with CCD and PDB. 
[Covalent radii](https://github.com/forlilab/Meeko/blob/904f6a2a10ef457ae654782c959d4654233f2991/meeko/linked_rdkit_chorizo.py#L120) of elements Ca and Fe are uncommented and set to 0 to be consistent with other atoms. Their chem templates are updated accordingly. 

- Added phosphate capping group and make it a possible match

Phosphate is used as a placeholder for missing residues in the 5' end of DNA/RNA. Sometimes, it's registered in the same residue (therefore we have the X5p embedding variants) but sometimes it's registered to the next (missing) residue. Making it a possible match will allow the retention of this phosphate, instead of deleting as a bad residue.

rwxayheee:
In https://github.com/forlilab/Meeko/pull/176/commits/6c807532933c0394c2338bc52ce3cf5a73f82cfe: The capping phosphate is dropped from ambiguous because it causes problems similar to NME, ACE as in #180. Currently, the excess phosphate, if recorded as a partially resolved residue in PDB, will be removed as bad residue. 

The chem templates are ready to merge ^^

diogomart (approved):

PR #183, opened 2024-10-01
 Fix delete_residues in chorizo (polymer) and mk_prepare_receptor

diogomart:
The list of residue IDs to delete is no longer passed to the `__init__` of the chorizo object. Instead, it lives only on the constructors `from_prody` and `from_pdb_string`. Then, the raw molecules passed to the `__init__` of the chorizo no longer contains deleted residues. In `mk_prepare_receptor`, deleted_residues are specified using the same syntax as flexres, e.g. `--delete_residues A:278`.


rwxayheee (approved):
This looks good. I tested with a few examples. Thanks so much!!

Issue #180, opened 2024-09-29
adjacent_res_smarts requires more atoms in adjacent_mol than capping residues (NME, ACE, phosphate) can provide


rwxayheee:
Hi, I was trying to register a phosphate capping group (similar to NME, ACE in proteins) as a new chem template. Just realized that this can be reproduced by protein examples. 

To reproduce -
Protein structure [2G5Z](https://www.rcsb.org/structure/2G5Z) is a green fluorescent protein (GFP) variant. Chain A is part of the beta-barrel and finishes with [NME](https://www.rcsb.org/ligand/NME). 

As the input, a stripped PDB file for chain A: 
[2g5z_A.pdb.txt](https://github.com/user-attachments/files/17181381/2g5z_A.pdb.txt)

Command: 
```
mk_prepare_receptor.py --macromol 2g5z_A.pdb --skip_gpf -o 2g5z
```

Output:
```
Traceback (most recent call last):
  File "/Users/amyhe/micromamba/envs/mk_dev/bin/mk_prepare_receptor.py", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/Users/amyhe/Desktop/0_forks/Meeko/scripts/mk_prepare_receptor.py", line 424, in <module>
    chorizo = LinkedRDKitChorizo.from_prody(
  File "/Users/amyhe/Desktop/0_forks/Meeko/meeko/linked_rdkit_chorizo.py", line 798, in from_prody
    chorizo = cls(
  File "/Users/amyhe/Desktop/0_forks/Meeko/meeko/linked_rdkit_chorizo.py", line 699, in __init__
    padded_mols = self._build_padded_mols(self.residues, bonds, padders)
  File "/Users/amyhe/Desktop/0_forks/Meeko/meeko/linked_rdkit_chorizo.py", line 1253, in _build_padded_mols
    padded_mol, mapidx = padders[link_label](
  File "/Users/amyhe/Desktop/0_forks/Meeko/meeko/linked_rdkit_chorizo.py", line 1990, in __call__
    raise RuntimeError(
RuntimeError: length of hits must be 1, but it's 0
```

I think the error occurred when it's trying to add the "5-term" padding to the normal residue that's adjacent to NME, when trying to find expected `adjacent_res_smarts` O=C[C:11][N:12] from its `adjacent_mol` that is NME. 

[...]

PR #184, opened 2024-10-02
Fixing NME in #180

rwxayheee:
This is a working version with minimal fixes for NME in #180 without changing the data structure of chem templates. 

Specifically for the following two conditions, some adaptions are made to "rxn_smarts" and "adjacent_res_smarts": 
(1) Residue -(C-term)-> NME for the making of padded **NME**
and
(2) Residue <-(N-term)- NME for the making of padded Residue 

In condition (2), NME is being padded, but it doesn't have enough atoms as required by the default reaction: 
```
"N-term": {
	    "rxn_smarts": "[C:1](=[O:2])[C:3][N:4]>>[C:1](=[O:2])[C:3][N:4][C:11](=[O:12])[C:13]",
	    "adjacent_res_smarts": "[C:11](=[O:12])[C:13][N]"
	},
```
Therefore, a new link label pointing to a different reaction is needed. 

In condition (1), NME is the next residue (**adjacent_mol**). https://github.com/rwxayheee/Meeko/commit/a242725ce16596c57a434f3a41d24896ac7a53dd has NME as a hardcoded exception. 

In the long term, if nonstandard residues with modified backbones are to be incorporated, it may be helpful to determine the best reaction by (the input resnames, or the categories of) both the target residue and the adjacent residue. 

Some refactoring of chem templates is underway to categorize residues and reactions, to handle padder building (and maybe also intermol bond perception) conditionally. If there are any updates they will be added to this PR.

### DIFF
--- a/meeko/data/residue_chem_templates.json
+++ b/meeko/data/residue_chem_templates.json
@@ -44,6 +44,10 @@
 	    "rxn_smarts": "[C:1](=[O:2])[C:3][N:4]>>[C:1](=[O:2])[C:3][N:4][C:11](=[O:12])[C:13]",
 	    "adjacent_res_smarts": "[C:11](=[O:12])[C:13][N]"
 	},
+	"NME-N-term": {
+	    "rxn_smarts": "[C:3][N:4]>>[C:3][N:4][C:11](=[O:12])[C:13]",
+	    "adjacent_res_smarts": "[C:11](=[O:12])[C:13][N]"
+	},
 	"C-term": {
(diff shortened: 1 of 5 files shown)


rwxayheee:
[e9bbc65](https://github.com/forlilab/Meeko/pull/184/commits/e9bbc6502ed39da4e80e3b1b118fcae9c0cf0014) will fail because of multiple matches when N-term linker is called for PRO. It's generally a bad idea to reduce number of atoms in adj mol (i.e. specificity of rxn) or to make further restrictions that are tied to residue names. I might be closing this once I submit the alternate solution.

PR #186, opened 2024-10-04
Alternate Locations

diogomart:


diogomart:
Still need to update the Prody path, but it is already working for `--pdb` / `LinkedRDKitChorizo.from_pdb_string()`.

rwxayheee (commented):
Hi @diogomart 
Thanks for the nice work. I tested some examples and it works for --pdb. I just have some very minor edits to protect from invalid inputs (empty or altloc with just space) that lead to the programming bug. Please let me know if I can be of any further assistance!

diogomart:
Great catch @rwxayheee thank you for testing!

rwxayheee (approved):
All looks good to me. I just added the missing ambiguous dict for ALA I noticed when testing. 
Please merge whenever you're ready to. Thanks again for the nice work!

Issue #142, opened 2024-07-26
NameError Encountered as Using `mk_prepare_ligand.py` with Argument `--multimol_prefix`


proline8314:
The following line is used to submit docking ligands handling jobs with Meeko.
```
conda activate <env name>
mk_prepare_ligand.py -i $1 --multimol_outdir $2 --multimol_prefix $3
```

And an error message occurs as follows.
```
File ".../miniconda3/envs/<env name>/lib/python3.10/site-packages/meeko/cli/mk_prepare_ligand.py", line 216, in __call__
2004     
name = '%s-%d' % (args.multimol_prefix, self.counter)
NameError: name 'args' is not defined
```

To address the issue, replacing `args` with `self` in line 216 appears to be a viable solution. Since the local variable `args` is not defined in the function `__call__`, and the previously defined class variable `self.multimol_prefix` seems to be the correct one to use.

diogomart:
Thanks, good catch! You have meeko v0.5.1, right?

proline8314:
Hi @diogomart 
Thank you for pointing that out. And yes, I believe the version is v0.5.1. I've also verified that the lastest release on Github has an identical line 216 for the file `mk_prepare_ligand.py` with the one I use.

diogomart:
Possibly caused by moving command line scripts  from `scripts/` to `meeko/cli/`, need to keep in mind as branch `develop` is merged onto `release` for v0.6.0

rwxayheee:
Thank you @proline8314 for the suggestion. We included this in #187

Issue #64, opened 2023-08-14
mk_prepare_ligand.py should return a non-0 exit code if it failed to create the output


xrobin:
Running mk_prepare_ligand.py on some unsupported ligands like [Sr+2] fails. This is fine, however the script returns a success exit code (0) which makes it difficult to detect the failure.

```
echo "
     RDKit          3D

  1  0  0  0  0  0  0  0  0  0999 V2000
    0.0000    0.0000    0.0000 Sr  0  0  0  0  0 15  0  0  0  0  0  0
M  CHG  1   1   2
M  END
$$$$
" > test.sdf
mk_prepare_ligand.py -i test.sdf -o test.pdbqt
```
The script fails and prints the following output:
```
[16:04:31] Warning: molecule is tagged as 3D, but all Z coords are zero
[16:04:31] ERROR: Problems encountered parsing data fields
[16:04:31] ERROR: moving to the beginning of the next molecule
atom number 0 has None type, mol name: 

```
However it returns a status code of 0 which makes it difficult to detect that something failed:
```
echo $?
0

```
It would be very nice if mk_prepare_ligand.py could return a non-0 exit code to indicate a failure occurred.

rwxayheee:
```
[16:04:31] Warning: molecule is tagged as 3D, but all Z coords are zero
[16:04:31] ERROR: Problems encountered parsing data fields
[16:04:31] ERROR: moving to the beginning of the next molecule
```
These are RDKit warnings or errors that may not immediately abort the execution (of reading a potentially multi-mol SDF) and do not necessarily lead to a failure to generate the PDBQT file. 

```
atom number 0 has None type, mol name: 
```
The above came from an unsuccessful attempt of PDBQTWriterLegacy.write_string. There's no immediate exit either, but the encounters are logged. 

It's understood that assignment of status code can be important, if there're external programs that depend on the exit code of mk_prepare_ligand.py. Proposed change: https://github.com/rwxayheee/Meeko/commit/7dcb9b111d5cd73adc0d59f4e64ddccab58669aa
```py
    # Determine if exit code should be non-zero based on processing results
    if output.num_files_written == 0:
        print("No PDBQT files were written due to errors!")
        sys.exit(1)  # full failure
    elif input_mol_with_failure > 0:
        print("Some molecules encountered errors!")
        sys.exit(2)  # partial failure
```

diogomart:
[...]

Issue #50, opened 2023-06-27
Missing name line in REMARK lines


gkxiao:
<p>The name of a molecule is a very important information. But the output file of meeko did not include the name line.</p>
<p>test.sdf:  an example of input file with molecular name "test"</p>
<pre line="1" lang="python">
test
 OpenBabel06272315433D

  4  3  0  0  0  0  0  0  0  0999 V2000
    1.0507    0.0872   -0.0485 C   0  0  0  0  0  0  0  0  0  0  0  0
    2.2754    0.0872   -0.0485 O   0  0  0  0  0  0  0  0  0  0  0  0
    0.4632    0.7434   -0.7103 H   0  0  0  0  0  0  0  0  0  0  0  0
    0.4632   -0.5690    0.6133 H   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  2  0  0  0  0
  1  3  1  0  0  0  0
  1  4  1  0  0  0  0
M  END
$$$$
</pre>
<p>test.pdbqt: the output PDBQT file prepared with meeko command:</p>
<pre line="1" lang="python">
mk_prepare_ligand.py -i test.sdf -o test.pdbqt
</pre>
<p>Here is the output.</p>
<pre line="1" lang="python">
REMARK SMILES C=O
REMARK SMILES IDX 1 1 2 2
REMARK H PARENT
REMARK Flexibility Score: inf
ROOT
ATOM      1  C   UNL     1       1.051   0.087  -0.049  1.00  0.00     0.307 C
ATOM      2  O   UNL     1       2.275   0.087  -0.049  1.00  0.00    -0.307 OA
ENDROOT
TORSDOF 0
</pre>
<p>There is no name line in the REMARK lines. Here is an example with Name line:</p>
<pre line="1" lang="python">
REMARK  Name = test
REMARK  0 active torsions:
REMARK  status: ('A' for Active; 'I' for Inactive)
REMARK                            x       y       z     vdW  Elec       q    Type
REMARK                         _______ _______ _______ _____ _____    ______ ____
ROOT
ATOM      1  C   UNL     1       1.051   0.087  -0.049  0.00  0.00    +0.000 C
ATOM      2  O   UNL     1       2.275   0.087  -0.049  0.00  0.00    +0.000 OA
ENDROOT
TORSDOF 0
</pre>

gkxiao:
The sdf file generated by mk_export.py is also missing the molecule name.

rwxayheee:
[...]

PR #187, opened 2024-10-09
Small fixes for ligand preparation

rwxayheee:
This PR has some small fixes related to ligand preparation, aiming to close the following tickets: 

- Include the fix suggested by @proline8314 in #142. 
- As suggested in #64 and @diogomart , mk_prepare_ligand.py will return non-zero exit codes if it fails to generate some (or any at all) PDBQT files for some molecules. 
- To address the request to preserve the ligand name in #50, mk_export.py will perceive the ligand name from the source (PDBQT, DLG) filename as usual and write to SDF file. No change in REMARK lines.
- Very small additional protection to mk_prepare_ligand.py: ensure opened file closure; check if mol is None before calling mol.GetProp and such


diogomart (approved):

diogomart:
Thanks!

PR #191, opened 2024-10-11
fix typo in _aux_altloc_mol_build

rwxayheee:
Fixed small variable name typo in _aux_altloc_mol_build

Also flipped missed and needed here
https://github.com/forlilab/Meeko/blob/8257379fdac56ab27b9a17f609ea443acf2111a0/meeko/utils/rdkitutils.py#L190-L193


diogomart (approved):
Great catch

rwxayheee:
Thank you @diogomart  I will merge with 7b59498. Very very small change in variable name

PR #193, opened 2024-10-12
add help print to mk_prepare_receptor

diogomart:
Adds help message for users who are used to `-o` flag in mk_prepare_receptor in older versions. Before v0.6.0, `-o` sets the PDBQT output filename, but in v0.6.0, it sets a default output basename that `--write` options rely on.

If someone runs `-o output.pdbqt -j`, the `-j` flag triggers writing a file named "output.pdbqt.json", but no PDBQT files are written. The added help message, shown below, is printed if the output basename ends with ".pdbqt" and `--write_pdbqt` was not used.

```
PDBQT files were NOT written. Use -p/--write_pdbqt for that.
Note that -o/--output_basename just sets a default for --write flags
```

### DIFF
--- a/scripts/mk_prepare_receptor.py
+++ b/scripts/mk_prepare_receptor.py
@@ -496,7 +496,6 @@ if args.add_templates is not None:
         sys.exit(2)
 templates = ResidueChemTemplates.from_dict(res_chem_templates)
 
-print(f"{templates=}")
 # create chorizos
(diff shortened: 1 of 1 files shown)


rwxayheee (approved):
This looks good, thank you! I wanted to drop the print(f"{templates=}") line too, thanks for doing that. I approve the PR

PR #197, opened 2024-10-15
re-implement residue.is_flexres_atom, and minor fixes

diogomart:
Populate list to label each atom as movable in a flexible sidechain or not. Some code from writer.py still needs to be removed as it's redundant now.


diogomart:
Superseeded by #200. This PR (#197) will be merged when #200 is merged.

PR #198, opened 2024-10-15
suppress rdkit reaction atom map warning in residue padder

diogomart:
For the reaction run by ResiduePadders, it suppresses the following rdkit warnings:
```
[23:00:47] product atom-mapping number 11 not found in reactants.
[23:00:47] product atom-mapping number 12 not found in reactants.
[23:00:47] product atom-mapping number 11 not found in reactants.
[23:00:47] product atom-mapping number 12 not found in reactants.
[23:00:47] product atom-mapping number 13 not found in reactants.
```
these arise because the product has atom labels that the reactants don't, but those labels are added on purpose. The filter is removed immediately after the reaction. Since the logger is a singleton, it is not guaranteed that the same warning won't be suppressed when multiprocessing from some other reaction that happens to be running, but I think that's OK.

### DIFF
--- a/meeko/__init__.py
+++ b/meeko/__init__.py
@@ -50,6 +50,15 @@ from .openff_xml_parser import load_openff
 from .openff_xml_parser import get_openff_epsilon_sigma
 from .hydrate import Hydrate
 
+import logging
+from rdkit import rdBase
+rdkit_logger = logging.getLogger("rdkit")
+rdkit_logger.handlers[0].setLevel("WARNING")
+rdkit_logger.handlers[0].setFormatter(
+    logging.Formatter('[RDKit] %(levelname)s:%(message)s'),
+)
+rdBase.LogToPythonLogger()
+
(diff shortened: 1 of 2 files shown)


rwxayheee (approved):
Looks good to me. Tested with some basic examples. Thanks!

meeko/__init__.py
```diff
@@ -50,6 +50,15 @@
 from .openff_xml_parser import get_openff_epsilon_sigma
 from .hydrate import Hydrate
 
+import logging
+from rdkit import rdBase
+rdkit_logger = logging.getLogger("rdkit")
+rdkit_logger.handlers[0].setLevel("WARNING")
+rdkit_logger.handlers[0].setFormatter(
+    logging.Formatter('[RDKit] %(levelname)s:%(message)s'),
```
rwxayheee, line 58: Because it's in init, it will change the format of RDKit logging messages (severity >= warning) once meeko is imported. Do you want to enforce this?

diogomart: Just leaving a record of what we discussed: I don't know. It may be useful to know which messages come from RDKit.

PR #200, opened 2024-10-15
Export sidechain full receptor

diogomart:


diogomart:
Exporting full PDB with docked flexible sidechain positions requires a chorizo object. When using from command line `mk_export` it is passed as a JSON file with `-j/--read_json`.

Summary of changes:
 - re-implemented the code that populates `ChorizoResidue.is_flexres_atom`. The attribute existed but it wasn't being populated (probably lost during the summer refactoring efforts).
 - modified `LinkedRDKitChorizo.to_pdb()` to optionally take a dictionary as argument, where an arbitrary number of new positions can be passed. This is a dictionary of dictionaries, the outer keys are residue IDs, the inner keys are atom indices, the inner values are positions. For example, to update a single atom position of atom 5 (0-index) of residue 42 in chain A:  `new_positions = {"A:42": {5: (x, y, z)}}`
 - add INDEX MAP to flexible residue PDBQT, similar to SMILES INDEX MAP for ligands
 - if no INDEX MAP, the residue templates are used to partially match the moving sidechain atoms. Kudos to @rwxayheee for this idea that enables using this feature with many flexible dockings ran to date. The atom names (not atom types) are used to infer the element.

meeko/__init__.py
```diff
@@ -40,6 +40,7 @@
 from .linked_rdkit_chorizo import ChorizoCreationError
 from .molecule_pdbqt import PDBQTMolecule
 from .rdkit_mol_create import RDKitMolCreate
+from .export_flexres import export_pdb_updated_flexres
```
rwxayheee, line 43: Both export_flexres and export_pdb_updated_flexres have been renamed or relocated..?

diogomart: Both renamed and relocated, e.g. `export_full()` in "linked_rdkit_chorizo.py", now called `export_pdb_updated_flexres()` in "export_flexres.py"

rwxayheee: export_flexres.py
Could you include the file in the PR?

rwxayheee: > export_flexres.py Could you include the file in the PR?

Could you do git add export_flexres.py? It seems like git isn't tracking it, I didn't see it in the branch. No hurry though, I can look at the other changes. I like get_updated_positions. it's reusable in many useful ways.

meeko/linked_rdkit_chorizo.py
```diff
@@ -343,6 +344,88 @@ def rectify_charges(q_list, net_charge=None, decimals=3) -> list[float]:
     return charges_dec
 
 
+def get_updated_positions(residue, new_positions: dict):
```
rwxayheee, line 347: (no rush to reply, just comment for record) I really like this. It works for dihedral rotor types of conformational changes. 
Is it safe to remove the `max one n2 absent from new_positions` restriction in condition A? If we're not restricting the number of n2, then this will also work for ring puckering.

rwxayheee: I really like `get_updated_positions`. I have following comments: 
1. Dropping `max one n2 absent from new_positions` might enable ring puckering. I don't 100% understand why the restriction was set for the condition, so maybe it's not safe to do so
2. Enumerating atoms in new_positions (not all atoms) and add hydrogens to set h_to_update might be slightly more efficient. 
3. n2 might not be heavy atom. I think hydrogens should be counted. When n2 is a hydrogen and a new position is being assigned, then the hydrogen needs to go to h_to_update.

diogomart: Excellent insight here, thanks.
1. Great point, I agree. I only added the restriction because I saw the hydrogen on a C-alpha get updated and I wished it kept the original position. But a better way to prevent the H on C-alpha from being updated is probably to remove the C-alpha from the `new_positions` argument, delegating the decision to the caller of `get_updated_positions()`, which feels right because that is where sidechain knowledge exists, and that sidechain knowledge is that C-alphas don't move because they are in the ROOT block of the PDBQT.

Pont 3. Agree.

Will implement 1. and 3. I didn't understand what you meant with 2., could you rephrase?

rwxayheee:
Commenting for record: 

[700ed89](https://github.com/forlilab/Meeko/pull/200/commits/700ed89b8e3b2c1668b8699db730738e4bdeb23d) has a rewrite of **get_updated_positions** on how `h_to_update` (list of hydrogen atom idx) is obtained. It drops the restrictions and only considers updating two general types of hydrogens: 
(A) need the update of position because parent heavy atoms (n1) has a new position
(B) need to undergo induced re-arrangement because the parent's neighbor (n2) has a new position

The code is written in a slightly different way. It iterates atoms with new positions, and update:
(A) the child hydrogens and 
(B) their neighbors' hydrogens

I didn't test the function yet, but will make sure to fix if there're any related problems.

diogomart:
I think all is implemented now and is ready to go

rwxayheee (approved):

rwxayheee:
Hi @diogomart 
Thanks again for the nice work. I approve the changes. 
Just one thing I wanted to check with you: 
Vina does not keep this line in the output
```
REMARK INDEX MAP 3 1 15 2 17 3 18 4 22 5
```
because it appears under BEGIN_RES
https://github.com/ccsb-scripps/AutoDock-Vina/blob/09f67a96edd3ba0c53608a3148fb2749d73635a9/src/lib/parse_pdbqt.cpp#L508-L518

I put the wrong link the first time, sorry

Do you want a small update on Vina to keep it? I can PR to AutoDock Vina

diogomart:
Thanks!

I ran with vina and it kept the INDEX MAP. Which version of vina did you use? I think I used 1.2.5. The vina output pdbqt file is in the tests: https://github.com/forlilab/Meeko/blob/f838e9dff37983320cfdf1b83c985187ec53de6e/test/linked_rdkit_chorizo_data/just-three-residues_vina_flexres_idxmap.pdbqt#L26

rwxayheee:
My bad. It does keep the line!! I have an overwrite protect where I run the tests. Never mind then. Please merge. Thanks again for the nice work

diogomart:
Thank for double checking that!

PR #204, opened 2024-10-17
Allow template fetching 

rwxayheee:
This PR includes the utilities for residue template generation. **Currently it works for noncovalent cofactors only. ** 

Here's a summary of the changes: 

- Insert a check for unknown residues in the `__init__` of `LinkedRDKitChorizo`. 
If `raw_input_mols` or `set_templates` contain residues not in the current loaded residue template, a new residue template will be made. 

- Fetch from rcsb
The `input_resname` or specified resname will be used to fetch a definition CIF file from RCSB (https://files.rcsb.org/ligands/download/). The CIF files will be downloaded as temporary files. 

- Build process
The chemical component will be processed into the canonical protonation state (after deprotonating a default set of acidic protons). No further embedding, extension, capping, linker guessing will be attempted. The logging level of the build process is controlled by `ChemicalComponent_LoggingControler`. Other associated functions are imported from `chemtempgen.py`. 

- Added dependencies
**gemmi**, copy, urllib.request (and network connectivity), time, tempfile

- Example output
```
(mk_dev) amyhe@Amys-MBP Downloads % mk_prepare_receptor.py -i 1d3g_input/1d3g.pdb -o 1d3g_output -j --default_altloc A
/Users/amyhe/micromamba/envs/mk_dev/bin/mk_prepare_receptor.py:4: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  __import__('pkg_resources').require('meeko==0.6.0a3')
@> 3144 atoms and 1 coordinate set(s) were parsed in 0.03s.
DEBUG:.prody:3144 atoms and 1 coordinate set(s) were parsed in 0.03s.
Input residue names {'DDQ', 'BRE', 'ORO', 'ACT', 'FMN', 'SO4'} not in residue_templates

WARNING:root:Molecule doesn't contain matching  with acidic_proton_loc {'[H][O][CX3](=O)': 0, '[H][O][SX4](=O)': 0, '[H][O][SX3](=O)': 0, '[H][O][PX4](=O)': 0, '[H][O][PX3](=O)': 0, '[H][O][CX3](=S)': 0, '[H][O][SX4](=S)': 0, '[H][SX2][a]': 0} -> deprotonate returning original mol... 
******************** New Template Built ********************
{
    "ambiguous": {},
    "residue_templates": {
        "DDQ": {
[...]

PR #206, opened 2024-10-21
Guess linker

rwxayheee:
This PR may supersede #204. It handles unbound cofactors and modified AA and NA linking fragements if no backbone modifications occur. Example output: 

```
(mk_dev) amyhe@Amys-MBP Downloads % mk_prepare_receptor.py -i 1iv8.pdb -o 1iv8_output -j 
/Users/amyhe/micromamba/envs/mk_dev/bin/mk_prepare_receptor.py:4: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  __import__('pkg_resources').require('meeko==0.6.0a3')
@> 6379 atoms and 1 coordinate set(s) were parsed in 0.03s.
DEBUG:.prody:6379 atoms and 1 coordinate set(s) were parsed in 0.03s.
Input residues {'A:27': 'MLZ', 'A:105': 'MLZ', 'A:108': 'MLZ', 'A:109': 'MLY', 'A:111': 'MLY', 'A:206': 'MLY', 'A:216': 'MLY', 'A:250': 'MLY', 'A:310': 'MLZ', 'A:326': 'MLZ', 'A:390': 'MLY', 'A:467': 'MLZ', 'A:504': 'MLY', 'A:508': 'MLZ', 'A:513': 'MLZ', 'A:519': 'MLZ', 'A:558': 'MLY', 'A:616': 'MLY', 'A:638': 'MLZ'} not in residue_templates

Trying to resolve unknown residues by building chemical templates... 
WARNING:root:Atom # 0 (N) in mol doesn't have implicit Hs -> continue with next atom... 
WARNING:root:Molecule doesn't contain wanted_smarts: [NX3h1] -> continue with next pattern... 
WARNING:root:Molecule doesn't contain pattern: [NX3h1] -> linker label for N-term will not be made. 
WARNING:root:Molecule doesn't contain wanted_smarts: [CX3h1] -> continue with next pattern... 
WARNING:root:Molecule doesn't contain pattern: [CX3h1] -> linker label for C-term will not be made. 
******************** New Template Built ********************
{
    "ambiguous": {
        "MLZ": ["MLZ_"]
    },
    "residue_templates": {
        "MLZ_": {
            "smiles": "[H]NC([H])(C=O)C([H])([H])C([H])([H])C([H])([H])C([H])([H])N([H])C([H])([H])[H]",
            "atom_name": ["H2", "N", "CA", "HA", "C", "O", "CB", "HB2", "HB3", "CG", "HG2", "HG3", "CD", "HD2", "HD3", "CE", "HE2", "HE3", "NZ", "HZ", "CM", "HCM1", "HCM2", "HCM3"],
            "link_labels": {"1": "N-term", "4": "C-term"}
        }
    }
}
************************************************************
******************** New Template Built ********************
{
    "ambiguous": {
        "MLZ": ["MLZ_N"]
    },
    "residue_templates": {
        "MLZ_N": {
            "smiles": "[H]N(C([H])([H])[H])C([H])([H])C([H])([H])C([H])([H])C([H])([H])C([H])(C=O)N([H])[H]",
            "atom_name": ["HZ", "NZ", "CM", "HCM1", "HCM2", "HCM3", "CE", "HE2", "HE3", "CD", "HD2", "HD3", "CG", "HG2", "HG3", "CB", "HB2", "HB3", "CA", "HA", "C", "O", "N", "H", "H2"],
            "link_labels": {"20": "C-term"}
        }
    }
}
************************************************************
******************** New Template Built ********************
{
    "ambiguous": {
        "MLZ": ["MLZ_C"]
    },
    "residue_templates": {
        "MLZ_C": {
            "smiles": "[H]NC([H])(C(=O)[O-])C([H])([H])C([H])([H])C([H])([H])C([H])([H])N([H])C([H])([H])[H]",
            "atom_name": ["H2", "N", "CA", "HA", "C", "O", "OXT", "CB", "HB2", "HB3", "CG", "HG2", "HG3", "CD", "HD2", "HD3", "CE", "HE2", "HE3", "NZ", "HZ", "CM", "HCM1", "HCM2", "HCM3"],
            "link_labels": {"1": "N-term"}
        }
    }
}
************************************************************
WARNING:root:Atom # 0 (N) in mol doesn't have implicit Hs -> continue with next atom... 
WARNING:root:Molecule doesn't contain wanted_smarts: [NX3h1] -> continue with next pattern... 
WARNING:root:Molecule doesn't contain pattern: [NX3h1] -> linker label for N-term will not be made. 
WARNING:root:Molecule doesn't contain wanted_smarts: [CX3h1] -> continue with next pattern... 
WARNING:root:Molecule doesn't contain pattern: [CX3h1] -> linker label for C-term will not be made. 
******************** New Template Built ********************
{
    "ambiguous": {
        "MLY": ["MLY_"]
    },
    "residue_templates": {
        "MLY_": {
            "smiles": "[H]NC([H])(C=O)C([H])([H])C([H])([H])C([H])([H])C([H])([H])N(C([H])([H])[H])C([H])([H])[H]",
            "atom_name": ["H2", "N", "CA", "HA", "C", "O", "CB", "HB2", "HB3", "CG", "HG2", "HG3", "CD", "HD2", "HD3", "CE", "HE2", "HE3", "NZ", "CH1", "HH11", "HH12", "HH13", "CH2", "HH21", "HH22", "HH23"],
            "link_labels": {"1": "N-term", "4": "C-term"}
        }
    }
}
************************************************************
******************** New Template Built ********************
{
    "ambiguous": {
        "MLY": ["MLY_N"]
    },
    "residue_templates": {
        "MLY_N": {
            "smiles": "[H]N([H])C([H])(C=O)C([H])([H])C([H])([H])C([H])([H])C([H])([H])N(C([H])([H])[H])C([H])([H])[H]",
            "atom_name": ["H", "N", "H2", "CA", "HA", "C", "O", "CB", "HB2", "HB3", "CG", "HG2", "HG3", "CD", "HD2", "HD3", "CE", "HE2", "HE3", "NZ", "CH1", "HH11", "HH12", "HH13", "CH2", "HH21", "HH22", "HH23"],
            "link_labels": {"5": "C-term"}
        }
    }
}
************************************************************
******************** New Template Built ********************
{
    "ambiguous": {
        "MLY": ["MLY_C"]
    },
    "residue_templates": {
        "MLY_C": {
            "smiles": "[H]NC([H])(C(=O)[O-])C([H])([H])C([H])([H])C([H])([H])C([H])([H])N(C([H])([H])[H])C([H])([H])[H]",
            "atom_name": ["H2", "N", "CA", "HA", "C", "O", "OXT", "CB", "HB2", "HB3", "CG", "HG2", "HG3", "CD", "HD2", "HD3", "CE", "HE2", "HE3", "NZ", "CH1", "HH11", "HH12", "HH13", "CH2", "HH21", "HH22", "HH23"],
            "link_labels": {"1": "N-term"}
        }
    }
}
************************************************************

Files written:
1iv8_output.json <-- parameterized receptor
```


rwxayheee:
[f464d7f](https://github.com/forlilab/Meeko/pull/206/commits/f464d7f4e7f2353bc444d67b0e808a1fa9964813) has a small fix for the problem in [comment](https://github.com/forlilab/Meeko/issues/207#issuecomment-2427972365)

When rdkit's ComputeGasteigerCharges returns NaN for unsupported elements, we will do 
```py
charges = [1e-6 if np.isnan(x) else x for x in charges]
```
as a walkaround. This is basically to put a 0. But [rectify_charges](https://github.com/forlilab/Meeko/blob/75305727874ced856d09dc41b697d97bf6194fb6/meeko/linked_rdkit_chorizo.py#L304) will redistribute the charges. It seems like, it can't be exactly 0. because of [divide_int_gracefully](https://github.com/forlilab/Meeko/blob/75305727874ced856d09dc41b697d97bf6194fb6/meeko/linked_rdkit_chorizo.py#L240)

This has been supersede by [e222e40](https://github.com/forlilab/Meeko/pull/206/commits/e222e4014276f38450fdca00929da1aff9f8ae3b) to handle inf

It's now
```py
charges = [1e-6 if np.isnan(x) or np.isinf(x) else x for x in charges]
```

rwxayheee:
[dde4d34](https://github.com/forlilab/Meeko/pull/206/commits/dde4d344cfbadcb36d5aca69782e3b5c066323ce) has a walkround for HEM and similar metal complexes (#207) without pre-registration of these templates. 

- Function recharge 
chemtempgen now processes supported metal complexes in two possible ways: 

(a) When the metal's charge state is ignored (= 0) in the definition file. 
---
In this case, we'll charge the coordinated atoms by the explicit valence (bond count) with the metal-nonmetal bonds, **assuming that the metal's charge has been transferred to the coordinated atoms**. 
The expected outcome will contain neutral metals and charged ligands. All metal-nonmetal bonds will be single bonds. 

(b) When all metals' charge states are explicitly assigned
---
The charge state will be taken as the metal's ionic charge, as if the complex were a salt and we'll **assume the bonds are ionized (fully polarized)**. For Fe, it's typically the oxidation state. 
In this case, the metal-nonmetal bonds will be polarized. In practice, the metal-nonmetal bonds are removed first. Next, the explicit valence of the coordinated atoms are re-calculated. Then, for each removed (polarized) bonds, the metal and coordinated atom are re-charged in pairs. 
The expected outcome may contain charged (and sometimes negatively charged) metals, which is a result of over-polarizing the bonds. All metal-nonmetal bonds will be single bonds. 

Although not chemically correct, it seems necessary to ensure all metal-nonmetal bonds are single bonds to be able to pass the template matching. I have tried BondType.ZERO but matching failed. 

So far it seems to work for some Fe- and Zn-containing ligands I tested.

rwxayheee:
This is ready. Commenting for record purpose: 

I consider the unspecified oxidation state a similar problem to the unspecified protonation / tautomeric state problem. The chemical templates generated on the run are always the best guess based on the definition. The proper way to parameterize the metal complexes is to **specify the total charge (this is possible through Python scripting) and perhaps also spin multiplicity**. But I think the guessing is useful because many times we don't know the exact charge state in the experimental structures. 

I gave up on my original plan for #207. It's impossible to make 100% correct templates from 99% incorrect definition files. There's no solution to guess the states. I would enumerate the oxidation states like we do with protonation states

diogomart:
This is great work. Found a couple an edge case:
- resname UNL results in a partially empty CIF that omits atom fields, triggering an error

And other issues:
- difficult to debug because of exceptions are caught and printed without the traceback. We can either reduce the code that is inside `try` statements, or re-raise the exceptions, or even consider something similar to this : https://github.com/forlilab/Ringtail/blob/4f43cb4fbbc50f04c6d3826c3b92692de08dc925/ringtail/logutils.py#L121-L137 
- temporary .cif files are not deleted

rwxayheee:
Hi @diogomart 

- UNL is a special case that doesn't have a concrete definition. It's usually a single unknown atom. 
If it presents the input, the only way to use it is to assign it or delete it. 
After ae38e53 + 086d5f0, UNL from input resname or assigned resname will be caught before template generation: 
```
(has_espaloma) amyhe@Amys-MBP Downloads % mk_prepare_receptor.py -i 1fsc.pdb -o 1fsc_output -j --set_template A:155=UNL
/Users/amyhe/micromamba/envs/has_espaloma/bin/mk_prepare_receptor.py:4: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  __import__('pkg_resources').require('meeko==0.6.0a3')
@> 521 atoms and 1 coordinate set(s) were parsed in 0.00s.
DEBUG:.prody:521 atoms and 1 coordinate set(s) were parsed in 0.00s.

Error: Creation of data structure for receptor failed.

Details:
Input residues {'A:156': 'UNL', 'A:157': 'UNL'} do not have a concrete definition
Input residues {'A:155': 'UNL'} do not have a concrete definition
Resdiues that are named UNL can't be parameterized. 

Recommendations:
1. (to parameterize the residues) Use --set_template to specify valid residue names, 
2. (to skip the residues) Use --delete_residues to ignore them. Residues will be deleted from the prepared receptor. 
```

- Difficult to debug 
Yes it might help to print Traceback. 
After 7b8f237 + ae38e53, now all traceback should be explicitly printed, if not None. 
```
(has_espaloma) amyhe@Amys-MBP ~ % mk_prepare_receptor.py -i 2kye.pdb -o 2kye_output -j
/Users/amyhe/micromamba/envs/has_espaloma/bin/mk_prepare_receptor.py:4: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  __import__('pkg_resources').require('meeko==0.6.0a3')
@> 480 atoms and 20 coordinate set(s) were parsed in 0.01s.
DEBUG:.prody:480 atoms and 20 coordinate set(s) were parsed in 0.01s.
Input residues {'A:6': 'PSU', 'A:7': 'PSU', 'A:12': 'PSU', 'A:14': 'PSU'} not in residue_templates

Trying to resolve unknown residues by building chemical templates... 

Error: Creation of data structure for receptor failed.

Details:
Max retries reached. Could not download CIF file for PSU. Error: <urlopen error [Errno 8] nodename nor servname provided, or not known>
  File "/Users/amyhe/micromamba/envs/has_espaloma/lib/python3.10/urllib/request.py", line 1348, in do_open
    h.request(req.get_method(), req.selector, req.data, headers,
  File "/Users/amyhe/micromamba/envs/has_espaloma/lib/python3.10/http/client.py", line 1283, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/Users/amyhe/micromamba/envs/has_espaloma/lib/python3.10/http/client.py", line 1329, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
 blablabla
```

- Temp files are usually cleaned up by system after a certain period of time. 
It might be more of a personal preference: When I develop similar codes, I would keep the files in /tmp just in case I want to take a look. The single cif files are not big. But we can clean up before exit if you like that better. 

After deb31c6
The temp files are deleted at exit. 

Let me know what you think! Thanks again for taking the time to look at it

diogomart:
Thanks, that looks great.
What prompted you to guard against Nan charges? https://github.com/rwxayheee/Meeko/commit/e222e4014276f38450fdca00929da1aff9f8ae3b
I think we should either raise an error or keep the nans, because zeroes make it look like everything went OK. https://github.com/forlilab/Meeko/pull/92

meeko/preparation.py
```diff
@@ -170,7 +170,7 @@ def __init__(
         self.dihedral_model = dihedral_model
         self.dihedral_params = dihedral_list
 
-        if dihedral_model == "espaloma" or charge_model == "espaloma":
+        if dihedral_model == "espaloma" or charge_model in ["espaloma", "zero"]:
```
diogomart: Why load espaloma for setting charges to zero?

rwxayheee: I wanted to say It throws an attribute not defined error otherwise (so sorry I accidentally edited the comment instead of replying to it

diogomart: Espaloma is imported inside the init of EspalomaTyper to prevent loading espaloma unnecessarily, because it takes a few seconds to load, and because it is an optional dependency. So. it would be better to set the attribute to None or similar

rwxayheee: does "zero" mean all zeros, or total charge = zero?

rwxayheee:
Hi @diogomart 
RDKit's ComputeGasteigerCharges might return NaN for metal complexes, and inf from strained geometries. 
Using espaloma will allow the same structures go through, although it shouldn't confidently output the charges with metals or strained geometries. 
I wanted the Gasteiger Charges to be forgiving because seems like espaloma is forgiving, but I can output errors if you think we should. Let me know which way you prefer!

meeko/preparation.py
rwxayheee: I thought "zero" still needs espaloma (or a charge method), the total charge is adjusted to zero, but maybe I'm wrong..

diogomart: It means all zeros. Should rename string to "all_zero" or similar.

I couldn't reproduce the attribute not defined error, but I tried only with ligand preparation. How did you get that? Was it with receptor prep?

rwxayheee: It means all zeros. Should rename string to "all_zero" or similar.
Got it! When I did this, I don't remember seeing charge list is set to all zero (or maybe it's initialized as all zeros? ), but I will take a closer look at the codes a little later this afternoon.

Yes, I found it with receptor prep. If I set MoleculePreparation(charge_model="zero") in the command line script

diogomart:
Need to guard for when `embed_allowed_smarts=None`.
Had a phosphorilated SER resnamed "T1P", which is a standalone molecule (not a linked one) in the component dictionary, and then got an error from rdkit trying to make a molecule from `embed_allowed_smarts`

rwxayheee:
This makes sense. I will fix it. The standalone molecules (no inter mol bonds) shouldn’t be evaluated for embedding

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

rwxayheee:
Hi @diogomart 
"T1P" is a modified nucleotide (maybe???). It's a possible linking fragment, but because of the nonstandard backbone it's not supported. The expected error is: 
```
Generation of chemical templates with modified backbones, which involves guessing of linker positions and types, are not currently supported. 
```
Could you please share the PDB ID? I want to see how it's linked in a polymer.

rwxayheee:
Also maybe you see a phosphorylated serine because Amber's residue name, but not the CCD residue name we would expect from PDB. 

I could add the Amber - CCD mapping, but I might need to screen Amber's definition first. It will take a little bit more time. 

Do you think it's a good idea to allow Amber resname? If so, I can do that

(I remember Amber also use CIF definition files). I can scrap the Amber24 residue library.

diogomart:
Just pushed the code to hold the metal aside during Gasteiger charge calculation. Tested with HEM and FDD. For FDD, final total charge is +5, because Fe comes in with +3 formal charge, and porphyrin with +2. How should we handle it?

rwxayheee:
Hi @diogomart 
FDD's definition is wrong. If the free base porphyrin is charged with +2, then it's incorrect to charge Fe by +3. This is why I guessed charges using method (b) in https://github.com/forlilab/Meeko/pull/206#issuecomment-2428874012. 

The template should have correct total charge. 

I can look at how you redistribute the charge. Will comment again in some moments

rwxayheee:
Hi @diogomart 

I have some quick explanation here. But if this still doesn't make sense, it might be better to discuss in person when we meet tomorrow, if that works for you. 
```py
(has_espaloma) amyhe@Amys-MBP Meeko % python
Python 3.10.15 | packaged by conda-forge | (main, Oct 16 2024, 01:24:20) [Clang 17.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from meeko.chemtempgen import *
>>> fdd = build_noncovalent_CC('FDD')
WARNING:root:Molecule contains metal specified charge state -> ionizing metal-nonmetal coordinate bonds...
WARNING:root:All metal-nonmetal coordinate bonds will be polarized. 
WARNING:root:Total charge of the molecule after recharging: 1
>>> fdd_from_smiles = Chem.MolFromSmiles(fdd.smiles_exh)
>>> Chem.rdmolops.GetFormalCharge(fdd_from_smiles)
-1
```

The total charge of FDD in template is -1. Here's a breakdown (salt model)
2x deprotonated carboxylate ... -2 
2x Ligand nitrogen atoms with incomplete valence (each has only two bonds) ... -2
Fe (III) ... +3

Using the remove-and-put-back strategy, we're extracting a neutral Fe, which means in the free base we want: 
2x deprotonated carboxylate ... -2 
an additional H+ (can be added to either of the nitrogen that were neutral) .. +1
This is because after restoring the valence, the 2x ligand nitrogen atoms with incomplete valence are neutralized now. 

The charge of Fe (III) in my understanding could be the sum of the 3 hydrogens (2 were used to complete the nitrogens, 1 was added to keep the total charge unchanged). The additional H+ is here to make up the difference between Fe (II) and Fe (III). For HEM with Fe(II), I feel that there won't be this additional H+

diogomart:
> FDD's definition is wrong. 

Awesome! Then +5 is expected behavior, as the method is preserving the total charge in an RDKit molecule that is passed **directly** to the MoleculeSetup constructor (i.e. without the fixes you implemented).

rwxayheee:
Hi @diogomart 

My current code generates this for FDD
```
[H]C1=C2C(C([H])([H])[H])=C(C([H])([H])C([H])([H])C(=O)[O-])C3=C([H])C4=[N+]5C(=C([H])C6=C(C([H])([H])[H])C(C([H])([H])[H])=C7C([H])=C8C(C([H])([H])[H])=C(C([H])([H])[H])C1=[N+]8[Fe-]5(N23)N76)C(C([H])([H])[H])=C4C([H])([H])C([H])([H])C(=O)[O-]
```

You can get the Smiles by: 
```py
from meeko.chemtempgen import *
fdd = build_noncovalent_CC('FDD') # this just needs to be a CCD ligand name
fdd.smiles_exh
``` 

which might lead to `expected only 1 added H per heavy atom`. This is likely because of [Fe-]. Therefore I think adding H+ to the free base is necessary, and this makes Fe(III) different from Fe(II). I can take a closer look later tonight. 

Without changing the code you added, it's also possible if I can neutralize the metal in Smiles. I'm not 100% sure if I can do this without breaking the single bonds, which are needed for template matching. But I will think about it

Thanks again for looking into this and for the nice work!

rwxayheee:
Hi @diogomart 

I tested the most recent version. The total charge seems correct for the ligands. There are just two minor things slightly different from what I thought, but I'm 100% fine with the current codes: 

1. Regardless of Fe(II) Fe(III) it seems like we're always adding four hydrogens to the free base. The protonated free base of FDD has a total charge of 0 (I thought it would be -1, because Fe is +3; see [the breakdown I commented above](https://github.com/forlilab/Meeko/pull/206#issuecomment-2436435430)). But it's not very important as long as the total charge of the reconstructed ligand is correct

3. It will not work for complexes with multiple metal centers. Complexes with multiple metal centers can go through the template generation (for example FES), but if we don't want to include them we can exclude at an earlier point. But they're very rare and if we don't want to do additional work, I'm also fine with the current error messages. 

I think it's ready, but happy to talk more if you would like to!! I really appreciate the great work, thanks a lot

rwxayheee:
I don't have any further edits. If both of us have tested it, please approve and we can merge. But if you have more edits, I will keep watching and very willing to test more. Thanks!

diogomart:
Weird thing:
```
logging.basicConfig(stream=sys.stdout, level=logging.WARNING)
```
resulted in prody's debug messages being shown, e.g.
```
DEBUG:.prody:2201 atoms and 1 coordinate set(s) were parsed in 0.27s.
```
changing the logging a bit, will push momentarily.

diogomart (approved):
This is a great feature. Thanks for all the work.
There might be further usability improvements, such as adding Amber resnames for CCD residues as you mentioned, but I think we should leave that for v0.7.

rwxayheee:
[eb6a69c](https://github.com/forlilab/Meeko/pull/206/commits/eb6a69c0800874864a267491e2bc7f4f2072cbbc) 
has a small bugfix for `bonded_unknown_res`

eb6a69c + 05e2d59 is a rewrite of how embed_allowed_smarts is handled. 
When embed_allowed_smarts is not specified (usage in Meeko), a guess for NA, AA will be made. A chemical component might match both. If so, chemical templates will be generated for all possible variants

rwxayheee:
I think it's ready to merge! ^^ Thank you @diogomart for the great work and kind support. I really appreciate it

PR #208, opened 2024-10-23
minor fix in error messages from mk_prepare_ligand stdout pdbqt

diogomart:
When using -/--, no PDBQT files are written, as expected, but there' was an unwanted error message about zero files being written.


rwxayheee (approved):
Great catch, thank you!

PR #211, opened 2024-10-24
Blunt end fix

rwxayheee:
Previously, the forgiving of bonded_atoms_missing in template matching allows residues nearby a gap to pass without manual specification of blunt ends. However, this is also causing multiple matches for CYX (disulfide-bond forming cystine): 

Input file: 
[1iep_receptorH.pdb](https://github.com/ccsb-scripps/AutoDock-Vina/blob/develop/example/basic_docking/data/1iep_receptorH.pdb)

```
(has_espaloma) amyhe@Amys-MBP output % mk_prepare_receptor.py -i 1iep_receptorH.pdb -o 1iep_receptor -p -v --box_size 20 20 20 --box_center 15.190 53.903 16.917
/Users/amyhe/micromamba/envs/has_espaloma/bin/mk_prepare_receptor.py:4: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  __import__('pkg_resources').require('meeko==0.6.0a3')
@> 4412 atoms and 1 coordinate set(s) were parsed in 0.02s.
DEBUG:.prody:4412 atoms and 1 coordinate set(s) were parsed in 0.02s.
ERROR:meeko.linked_rdkit_chorizo:for residue_key='A:305', 3 have passed
these are: ['CYS', 'CYX', 'CYX-']
and were evaluated for the number of missing Hhowever there was a tie between
ERROR:meeko.linked_rdkit_chorizo:for residue_key='A:330', 3 have passed
these are: ['CYS', 'CYX', 'CYX-']
and were evaluated for the number of missing Hhowever there was a tie between
ERROR:meeko.linked_rdkit_chorizo:for residue_key='A:369', 3 have passed
these are: ['CYS', 'CYX', 'CYX-']
and were evaluated for the number of missing Hhowever there was a tie between
ERROR:meeko.linked_rdkit_chorizo:for residue_key='A:464', 3 have passed
these are: ['CYS', 'CYX', 'CYX-']
and were evaluated for the number of missing Hhowever there was a tie between
ERROR:meeko.linked_rdkit_chorizo:for residue_key='A:475', 3 have passed
these are: ['CYS', 'CYX', 'CYX-']
and were evaluated for the number of missing Hhowever there was a tie between
WARNING:meeko.linked_rdkit_chorizo:No template matched for residue_key='A:498'
tried 3 templates for residue_key='A:498'excess_H_ok=False
GLN        heavy_miss=0 heavy_excess=0 H_excess=1 bond_miss={0} bond_excess=set()
NGLN       heavy_miss=0 heavy_excess=0 H_excess=1 bond_miss={0} bond_excess={4}
CGLN       heavy_miss=1 heavy_excess=0 H_excess=1 bond_miss=set() bond_excess=set()

  File "/Users/amyhe/micromamba/envs/has_espaloma/bin/mk_prepare_receptor.py", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/Users/amyhe/Desktop/0_forks/Meeko/scripts/mk_prepare_receptor.py", line 512, in <module>
    chorizo = LinkedRDKitChorizo.from_prody(
  File "/Users/amyhe/Desktop/0_forks/Meeko/meeko/linked_rdkit_chorizo.py", line 1040, in from_prody
    chorizo = cls(
  File "/Users/amyhe/Desktop/0_forks/Meeko/meeko/linked_rdkit_chorizo.py", line 877, in __init__
    padded_mols = self._build_padded_mols(self.residues, bonds, padders)
  File "/Users/amyhe/Desktop/0_forks/Meeko/meeko/linked_rdkit_chorizo.py", line 1568, in _build_padded_mols
    raise RuntimeError(err_msg)
RuntimeError: expected two paddings for ('A:305', 'A:306') (1, 4), padded 1
expected two paddings for ('A:330', 'A:331') (1, 4), padded 1
expected two paddings for ('A:369', 'A:370') (1, 4), padded 1
expected two paddings for ('A:464', 'A:465') (1, 4), padded 1
expected two paddings for ('A:475', 'A:476') (1, 4), padded 1
```

This PR makes the forgiving a fallback option, only effective when there's no match.


rwxayheee:
In the fallback block, I wanted to also allow H_excess. 

This will allow the C-terminal GLN residue, A:498, to pass as if it's a residue nearby a gap, with a match with GLN (embedded). 

This is probably a good description, because in the input structure GLN is capped by H so it ends with just C(=O)[H]. In the original structure, there're subsequent residues that went missing after GLN

The change is in 974f09e.

rwxayheee:
[974f09e](https://github.com/forlilab/Meeko/pull/211/commits/974f09e441914cf754a3165c73710b71bbb6f535) was a bad idea. I wanted to forgive excess H at linker atoms only.. but currently only the number of H_excess is counted.

rwxayheee:
A walkaround is changing H_excess from atom counts to a list of parent atoms' indices; with this, it's possible to determine whether H_excess occurs at link atoms: 
37239a8
But maybe it's not the best way to fix it.. TT

diogomart:
d7af3d277317930951052e00991087192b45d9b2 causes N-terminal residues to match NH3+ template instead of blunt neutral backbone residue, i.e. NMET vs MET for residue 15 in "test/linked_rdkit_chorizo_data/just-three-residues.pdb"

rwxayheee:
@diogomart 
I feel like this is a natural thing (unless we start to check H_missing), but i could be wrong :'(( For the N-terminal MET I got: 

```
MET        heavy_miss=0 heavy_excess=0 H_excess=[] bond_miss={4} bond_excess=set()

NMET       heavy_miss=0 heavy_excess=0 H_excess=[] bond_miss=set() bond_excess=set()

CMET       heavy_miss=1 heavy_excess=0 H_excess=[] bond_miss={5} bond_excess={1}
```

Was this expected? It seems like NMET will be prioritized based on the match results (regardless of how we forgive the match problems). I didn't change the way bond_miss was determined. 

I will think about this..

rwxayheee:
I have a solution (maybe not the best!!): 

(highest) priority: always match the embedded form (MET) first with tolerance of missing bonds at linkers
Normal MET, blunt MET (including C-term MET and N-term MET that are broken at the outer linker atoms) will come through as MET

(medium): only when we can't match with embedded, try matching terminal forms (CMET and NMET) with no excess forgiveness
Normal CMET and NMET will come through as CMET and NMET

If there's still no match, then there's a problem. I will try and see if there're any more exceptions

rwxayheee:
Hi @diogomart 
Could you please take a look at 
68ea823

The search of best match templates now has maximum of three rounds: 

- In the 1st round
residues that match exactly with the embedded templates (MET) will pass

- 2nd round (only happens when no match in 1st round)
residues that could be embedded will pass, when we forgive H_excess and bond_missing at linkers

- 3rd round (only happens when no match in 1st and 2nd rounds)
residues that are exactly the terminal (NMET, CMET) will pass

All tests + All Vina tutorial examples will pass. I still have some minor doubts on CYX (I think it needs 3 paddings), but some changes to bonds, count of bond used and number of paddings may be required.

diogomart:
@rwxayheee I think we should **not** forgive missing bonds for disulfide bridges. Thinking of adding an attribute to the padders that would state if a missing bond associated with the padder can be forgiven, e.g.:
```
	"5-prime": {
		"rxn_smarts": "[PX4h1:1]>>[P:1][O:11][C:12]",
		"adjacent_res_smarts": "[O+0X2h1:11][CX4:12]1CC(n)[OX2]C1",
		"auto_blunt": true
	},
```

Motivation 1:
The user should manually decide how to protonate CYS. 

Motivation 2:
The test system I have now is a CYX- at a C-terminal, so it's missing two bonds, and since both are forgiven at the second round of matching, and then I get a tie between the "CYX-" and "CYX" templates.

What do you think?

rwxayheee:
That’s a great point!! Yes, I agree it’s not safe to forgive linkers.
I was thinking about using a lookup table of scores to prioritize the candidate templates (and penalize the problems differently). It will slightly shorten the multiple if (forgive/not forgive) process I implement.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

rwxayheee:
Please forget about my suggestion for using scores. I was thinking because we're checking multiple factors and having a sum of the factors is straightforward. But at the moment the best solution is as you said including a new attribute for padders. I like the idea

Would you please implement "auto_blunt"? I can help check the implementation with some case specific tests. Thanks!

diogomart (approved):

Issue #210, opened 2024-10-24
Resolve residue name conflicts with Amber inputs by adding ambiguous definitions to default templates


rwxayheee:
This may not be happening for 0.6.0, but will not change the logics of current template matching. The changes will be just about the template files: 

- First, available Amber residue names will be added to default chemical template. Instead of finding the best match from CCD, they will be parsed directly from leap dat files. These names are the expected residue names from Amber compliant input. They will be the parents to lead a new item in the ambiguous dict, if they're never registered. 

- Then, CCD ligands whose names coincide with Amber residue names (HIE, T1P) will be added to default chemical template. These templates are added as possible matches for the residue names in the ambiguous dict.

rwxayheee:
Closing as #216 is merged

PR #216, opened 2024-10-28
Add residue templates to disambiguate Amber resname and CCD names

rwxayheee:
This is for #210. It only includes addition of chemical templates, and it does not change the matching of existing residue names. In this PR, 107 new ambiguous residue names and 567 unique templates are added to the default chemical template file, `residue_chem_templates.json`. It's also possible to distribute the new templates by libraries, or as a separate file. 

The technical details of the additional templates are as follows: 

- Disambiguation
The purpose of putting additional templates into the default template file is disambiguation in case of conflicts between Amber residue names and CCD names. All possible matches and the variants are registered under the same parent (an ambiguous residue name). 

- Source of additional residues
The following Amber OFF lib files in Amber24 are picked as source of additional residues: 
```py
leap_lib_path = '$AMBERHOME/dat/leap/lib/'
all_lib = (leap_lib_path+x for x in (
    # AA in ff19SB
    'amino19.lib',
    'amino19ipq_0.9.lib',
    'aminoct19ipq_0.9.lib',
    'aminont19ipq_0.9.lib',
    # mod AA
    'phosaa19SB.lib', 
    'mod_amino19.lib',
    # RNA
    'RNA.lib',
    'terminalphos.LJbb-RNA.lib',
    # DNA
    'DNA.OL15.lib',
    'parmBSC1.lib',
    # mod NA
    'all_modrna08.lib',
))
```
The outcomes are combined in a non-overwriting manner, as residues with duplicate names are skipped. 14 Residues with   unsupported elements and corrupt atomic numbers were discarded. 

- Processing of Amber residues
The lib files are parsed with ParmEd, and the rdkit molecule is created by a constructor in chemtempgen. It should be noted that **Amber OFF lib is not the ideal file type to generate a residue's Smiles**, as the valence and formal charge are not available from the files. A graph-based method is used to guess double bonds with some hints from atom types. Contrary to the processing of chemical components from CCD, no deprotonation occurred in the processing of Amber residues. 

- Suffix explained
```
        "NLE": ["NLE", "NLE_N", "NLE_C", "NLE_fl-ccd", "NLE_C-ccd"],
```
NLE: Residue NLE from Amber lib, forged into the embedded linking fragment
NLE_N: Residue NLE from Amber lib, as an N-term residue
NLE_C: Residue NLE from Amber lib, as a C-term residue. This variant usually has C(=O)[H], because residues from Amber usually do not have a full carboxylate/phosphate group. 
NLE_fl-ccd: Residue NLE from CCD, in the free-ligand form. No embedding was made
NLE_C-ccd: Residue NLE from CCD, as a C-term residue. This variant usually has C(=O)[O-]

### DIFF
--- a/meeko/data/residue_chem_templates.json
+++ b/meeko/data/residue_chem_templates.json
@@ -32,7 +32,114 @@
 		"THR": ["THR", "NTHR", "CTHR"],
 		"VAL": ["VAL", "NVAL", "CVAL"],
 		"TRP": ["TRP", "NTRP", "CTRP"],
-		"TYR": ["TYR", "NTYR", "CTYR"]
+		"TYR": ["TYR", "NTYR", "CTYR"],
+        "HYP": ["HYP", "HYP_fl-ccd"],
+        "MTB": ["MTB", "MTB_fl-ccd"],
+        "NLE": ["NLE", "NLE_N", "NLE_C", "NLE_fl-ccd", "NLE_C-ccd"],
+        "NHE": ["NHE", "NHE_fl-ccd"],
+        "S1P": ["S1P", "S1P_N", "S1P_C", "S1P_fl-ccd"],
+        "SEP": ["SEP", "SEP_N", "SEP_C", "SEP_fl-ccd", "SEP_C-ccd"],
+        "T1P": ["T1P", "T1P_N", "T1P_C", "T1P_fl-ccd"],
+        "TPO": ["TPO", "TPO_N", "TPO_C", "TPO_fl-ccd", "TPO_C-ccd"],
+        "Y1P": ["Y1P", "Y1P_N", "Y1P_C", "Y1P_fl-ccd"],
+        "PTR": ["PTR", "PTR_N", "PTR_C", "PTR_fl-ccd", "PTR_C-ccd"],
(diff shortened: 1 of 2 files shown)


diogomart:
This is awesome

diogomart (approved):

rwxayheee:
I will merge a little later today, after some visual inspection. Thanks for the approval! ^^

rwxayheee:
Some additional notes for future reference: 

Since bond order and formal charge can't be parsed from Amber OFF lib, it's not guaranteed that the Smiles strings in the chemical templates are the preferred/intended resonance form. For example, Amber residue 1MA: 
https://github.com/forlilab/Meeko/blob/ae23f949c642a17291972e70f779ea3d4ebfe8c3/meeko/data/residue_chem_templates.json#L1304-L1307
In the creation (guess) of conjugate bond system, a connected graph of atoms needing valence without changing its formal charge is considered, and the double bonds are first placed on the longest Eulerian path with an even number of edges. This is impossible when there are more than 2 odd (1-degree) nodes. 1MA is an example that contains a subgraph that has 3 odd nodes. The current compromise strategy is to remove the closest leaf node to a high-degree node. In 1MA, the valence of the removed node is conpensated by increasing the bond order with and upcharging the nitrogen in -NH2. This process doesn't pick a particular nitrogen. 

In short, it's understood that the Smiles could sometimes be alternatively written with a different resonance form. But this doesn't seem to really affect the matching we do in Meeko.

PR #218, opened 2024-10-30
use absolute import to import covalent_radius in chemtemgen.py

rwxayheee:
There was a small problem when I wanted to `import meeko.chemtempgen` in a Python script. Changing 

```py
from .utils.rdkitutils import covalent_radius
```

To

```py
from meeko.utils.rdkitutils import covalent_radius
```

### DIFF
--- a/meeko/chemtempgen.py
+++ b/meeko/chemtempgen.py
@@ -9,7 +9,7 @@ import re
 import atexit
 import os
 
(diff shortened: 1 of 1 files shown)


diogomart:
will be merged automatically with #219

rwxayheee:
Yes, I just realized that I included it in #219. Thanks a lot for looking into the PRs. I will close this one

PR #222, opened 2024-11-04
R.I.P. Chorizo

diogomart:
Rename:
- `LinkedRDKitChorizo` to `Polymer`
- `ChorizoResidue` to `Monomer`

Previously variable names were usually:
```python
for res_id, residue in chorizo.residues.items():
````

Now, the code looks like:
```python
for res_id, monomer in polymer.monomers.items():
```

Note that the key `res_id` is still called `res`. Similarly, `ResidueTemplate`, `ResiduePadder`, and `ResidueChemTemplates` are still called residues, not monomers. Only the `Monomer` objects (formerly `ChorizoResidue`) are referred to as monomers, everything else continues to be referred to as residues.


rwxayheee (approved):
This looks good, thank you for the hard work. I approve the PR. This is so much more work that I thought. I have to say I also like the less generic names that help prevent potential name collisions and improve readability, but I appreciate the work. Thank you @diogomart

PR #226, opened 2024-11-05
Prep v0.6.0

diogomart:
Updated version strings to v0.6.0
Moved scripts to `meeko/cli`, per conda-forge requirements and already implemented in v0.5.1
Removed script `mk_get_openff_epsilon_sigma.py`


rwxayheee (approved):
This looks good, thanks!

rwxayheee:
Would you want to put dependencies in setup.py `install_requires=`, or do you prefer them to stay in installation guide? 
I'm fine with either way
The template making function introduces a new dependency (gemmi)

diogomart:
I'm including as you suggest unless @nbruciaferri objects due to conflicts with conda.
Advantage of including is that a local pip install from source `pip install .` will automatically fetch the dependencies from PyPI. Numpy is already included, we are missing rdkit, scipy, and gemmi. Not including prody because it brings biopython, and pyparsing, and downgrades numpy, and it is possible that users have a more complex environment situation if they are installing from source. Users can always manually install prody. Also making pandas an optional dependency, only used to convert interaction analysis to a dataframe. This function is not used by ringtail.

rwxayheee:
Thanks! Ringtail also needs pandas although there's no explicit `install_requires=` in setup.py. I'm completely fine either way

diogomart:
I think a pip installation from PyPI should install dependencies. At least that's what I'm seeing for popular packages. For example, `pip install scipy` installs numpy. And pandas and matplotlib install a bunch of stuff. The issue for ringtail might be that chemicalite is not on PyPI, only on conda-forge, but @maylinnp might drop that dependency in the future. https://github.com/forlilab/Ringtail/issues/58

nbruciaferri:
As long as we already checked that the packages in the `install_requires=` are working and available I'm fine with it.

nbruciaferri (approved):
Looks good to me, ready to merge!

PR #238, opened 2024-11-12
fix link to readthedocs, fix cli help msg

diogomart:
README links to readthedocs now work

as suggested in https://github.com/ccsb-scripps/AutoDock-Vina/issues/319 help message for flexres option in mk_prepare_receptor.py now does not suggest that chain ID is optional, new message reads:
> leave chain ID empty if omitted in input PDB or mmCIF

old message was:
> skip chain ID if it is unspecified


rwxayheee (approved):
Thanks!

PR #241, opened 2024-11-14
add meeko/data/params/* to MANIFEST.in

diogomart:
it appears that basic functionality isn't working with meeko installed from pip or conda, because the JSON files with the default atom types isn't being distributed

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,5 +3,6 @@ include LICENSE
 include test/*
 include example/*
 include meeko/data/*
+include meeko/data/params/*
 include meeko/tmp/*
 include meeko/utils/*


rwxayheee (approved):
Thanks!

rwxayheee:
It seems like MANIFEST.in's include is not recursive. Would you like to include subdirectories in test/?

diogomart:
Good question. I don't know. It wouldn't hurt so I guess we could do it. The only scenario I can think in which that would be used is to check the installation, but having the user `cd` into the installation directory feels cumbersome. I can't find a way to run `pytest` from anywhere in the filesystem. But we should add those too, just in case.

PR #249, opened 2024-11-17
uniformize os.linesep as eol (no pathlib.os in Python 3.13)

diogomart:
thanks @atillack for pointing out that `pathlib.os` was dropped in Python 3.13

this PR uniformizes the use of `os.linesep` as `eol`

several unused `import os` were removed


rwxayheee (approved):
Looks good to me, thanks!

PR #254, opened 2024-11-27
add cycle_break attribute to molsetup Bond

diogomart:
This is to support a more general macrocyle bond breaking scheme in future autodock, which does not depend on atom types such as CG0/G0. Adds option `untyped_macrocycle` to `MoleculePreparation` to skip assigning the CG/G types, and to allow any rotatable bond to be broken for macrocycle sampling, instead of allowing only bonds between atoms of type C.


diogomart:
@rwxayheee do you like `Bond.breakable` ? I considered also `broken`, `cycle_break`, and `to_break`. Didn't go with `cycle_break` because it could be used beyond macrocycles. If it doesn't sound wrong to you, it's good enough.

rwxayheee (approved):
Hi @diogomart, The names and concepts make sense to me. Looks good to me, I approve the PR. Thank you!

diogomart:
Thank you

Issue #233, opened 2024-11-12
Small problem with delete_atom or molsetup.Atom


rwxayheee:
There might be a small problem with function `delete_atom` (making a dummy atom) in `MoleculeSetup`
https://github.com/forlilab/Meeko/blob/c938ebdc347bda69f0247dbd3c7566e8e27d1dea/meeko/molsetup.py#L687

```py
from meeko.molsetup import Atom
Atom(1, is_dummy=True)
```

None of our cli scripts are currently using this, but I can provide a fix in the future

rwxayheee:
fixed in #255

Issue #47, opened 2023-05-08
retaining unique atom names in the .pdbqt file


alfredoq:
hello,

when preparing a .pdbqt file I noticed that the unique atom name generated by RDKit are removed. An example below:

_from rdkit import Chem
from rdkit.Chem import AllChem_

_smiles = "CCCCO"
m = Chem.MolFromSmiles(smiles)
m3d = Chem.AddHs(m)    
AllChem.EmbedMolecule(m3d)
pdb_file = Chem.MolToPDBBlock(m3d)_

in this case _pdb_file_ contains:

_HETATM    1  C1  UNL     1      -1.515   0.808   0.093  1.00  0.00           C  
HETATM    2  C2  UNL     1      -0.734  -0.344   0.622  1.00  0.00           C  
HETATM    3  C3  UNL     1       0.290  -0.899  -0.311  1.00  0.00           C  
HETATM    4  C4  UNL     1       1.341   0.103  -0.717  1.00  0.00           C  
HETATM    5  O1  UNL     1       2.046   0.592   0.381  1.00  0.00           O  
HETATM    6  H1  UNL     1      -2.495   0.475  -0.327  1.00  0.00           H  
HETATM    7  H2  UNL     1      -1.772   1.505   0.929  1.00  0.00           H  
HETATM    8  H3  UNL     1      -1.014   1.393  -0.690  1.00  0.00           H  
HETATM    9  H4  UNL     1      -1.445  -1.161   0.864  1.00  0.00           H  
HETATM   10  H5  UNL     1      -0.203  -0.023   1.548  1.00  0.00           H  
HETATM   11  H6  UNL     1       0.838  -1.695   0.263  1.00  0.00           H  
HETATM   12  H7  UNL     1      -0.134  -1.413  -1.197  1.00  0.00           H  
HETATM   13  H8  UNL     1       2.046  -0.435  -1.397  1.00  0.00           H  
HETATM   14  H9  UNL     1       0.948   0.957  -1.266  1.00  0.00           H  
HETATM   15  H10 UNL     1       1.801   0.136   1.204  1.00  0.00           H_

with atoms having unique names. However after processing this RDKit molecule to get the corresponding .pdbqt file:

_preparator = MoleculePreparation()
preparator.prepare(m3d)
pdbqt_string = preparator.write_pdbqt_string()
pdbqt_string_ready = pdbqt_string.replace('UNL','MOL')   
ligand_file = '/home/fredy/Desktop/test.pdbqt'   
[...]

PR #255, opened 2024-11-27
Add keyword argument `rename_atoms` to `MoleculePreparation.prepare` and make option avail in CLI script 

rwxayheee:
This PR enables unique atom names in MoleculeSetup after MoleculePreparation with the relevant option. The main purpose is to address request #47. Option `--rename_atoms` is added to the CLI script in an analog to In `prepare_ligand4.py`; there's a `-w` option that renames ligand atoms, and `each newname is the original name of the atom plus its (1-based) uniqIndex`. 

Here're more details: 

- Add keyword argument `rename_atoms` to `MoleculePreparation.prepare`
The default is **False**, and it's just a keyword argument in function `prepare`. Because I didn't make it an attribute of MoleculePreparation, it can't be from the config (but we could, if the other way is preferred?). Python usage: 

```py
preparator = MoleculePreparation()
molsetups = preparator(mol, rename_atoms = True)
```

is equivalent to 

```py
molsetups = preparator.prepare(mol, rename_atoms = True)
```

- Make option available in `mk_prepare_ligand.py`
`mk_prepare_ligand.py` has a new option: 

```
  --rename_atoms        rename atoms: the new name will be the original name and its (1-based) index in
                        MoleculeSetup
```

- Format of new names
The new names are a combination of original name plus the atom's (1-based) index in MoleculeSetup at the last moment of molecule preparation. For example: `C ` might become `C14` and `OH` might become `OH15`. The index should be generally consistent with the order of atoms in the input molecule, but not Smiles or the order of output PDBQT file. And glue atoms will have indices too. When the name is longer than 4 characters, the original name will remain unchanged and a warning message shall be seen: 

```
Warning: Attempted to rename atom with original name XXX to XXX10. 
But the new name is too long (> 4 characters). The original name will be kept. 
```

- Assign/Import atom names
It's always possible to assign atom names in a rdkit molecule before MoleculePreparation. To import atom names, possible source includes: CIF files, MOL2/MOL files, PDB files, but PDB files are currently the only ones where atom names are parsed natively by RDKit. 

This started from the release branch, so it includes some other commits that have been merged into release but not present in develop. It also includes another small fix (14ec360) just for #233, although noncritical to the rest of this PR. 

@diogomart Please let me know what you think! Thanks for your time and advice in advance


diogomart (approved):
Thanks!

rwxayheee:
Thanks! Merging to develop

PR #257, opened 2024-11-30
Use utility functions in jsonutils to fix #151

### LINKED ISSUE #151: Can't serialize MoleculeSetup dihedrals to JSON
JSON doesn't like tuples for keys.

rwxayheee:
This is for #151. Used existing utility functions in [jsonutils](https://github.com/forlilab/Meeko/blob/release/meeko/utils/jsonutils.py), tuple_to_string and string_to_tuple to enable encoding and decoding of rotamers data of `rotamers` data for `MoleculeSetup` and `dihedral_partaking_atoms` and `dihedral_labels` for `RDKitMoleculeSetup`. 

@diogomart Could you please review the fixes and let me know if they work as expected? Additionally, I would like to become more familiar with parameter passing, if we might want to store Amber (and other) force field parameters in the future. This is not urgent, please feel free to advise at your convenience

### DIFF
--- a/meeko/molsetup.py
+++ b/meeko/molsetup.py
@@ -1377,7 +1377,7 @@ class MoleculeSetup:
             obj["ring_closure_info"]["bonds_removed"],
             obj["ring_closure_info"]["pseudos_by_atom"],
         )
-        molsetup.rotamers = obj["rotamers"]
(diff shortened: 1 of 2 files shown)


diogomart (approved):
Looks great, thanks! Added a test.

Issue #224, opened 2024-11-05
Keep partial charges from mol2 ligand file


tcaceresm:
Hi!
I converted a  mol2 file (with partial charges) to pdbqt file using meeko. It works, but partial charges are different than mol2 file. When I use openbabel to perform the file convertion, the partial charges of pdbqt are exactly the same than mol2 (with exception of C-H bonds, as expected).
Am I missing some option to keep partial charges from initial file in meeko?
Thanks!

diogomart:
Hello,
We don't have this feature at the moment. But it's possible we will implement it in the future. You can always use the Python API to get the partial charges from the RDKit molecule and copy them to the Atom instances in meeko's MoleculeSetup. If all you need is to keep the charges on the non-polar hydrogens, you can run mk_prepare_ligand with option `--merge_these_atom_types` and don't pass any values to this option, then hydrogens are not merged.

rwxayheee:
Hi @tcaceresm 
We have just implemented some way to take existing partial charges for ligand preparation, which includes MOL2 files. This feature is now available in the `develop` branch. 

Sample input file: 
[212163792.mol2.txt](https://github.com/user-attachments/files/18075234/212163792.mol2.txt)

Example usage 1 (with command-line script)

```
mk_prepare_ligand.py -i 212163792.mol2 --charge_model read
```

Example usage 2 (In Python)

```py
from meeko import MoleculePreparation
from meeko import PDBQTWriterLegacy
from rdkit import Chem

input_filename = "212163792.mol2"
output_filename = "212163792.pdbqt"
mol = Chem.MolFromMol2File(input_filename, removeHs=False)
mk_prep = MoleculePreparation(charge_model="read", charge_atom_prop="_TriposPartialCharge")
molsetup_list = mk_prep(mol)
molsetup = molsetup_list[0]
pdbqt_string = PDBQTWriterLegacy.write_string(molsetup)

with open(output_filename, "w") as f:
    f.write(pdbqt_string[0])
```

Please give it a try and let us know what you think ^^ feel free to comment or re-open this issue if we can be of further assistance

tcaceresm:
@rwxayheee hi! 
I tested it with a couple of molecules and it works well!
Thanks!

PR #258, opened 2024-11-30
Read Partial Charges from mol2 and SDF

rwxayheee:
This is for #224. 

`Chem.MolFromMol2File` in RDKit parses partial charges into atom property name `_TriposPartialCharge`. This PR simply reads the charges from atom property at the point [init_atom](https://github.com/forlilab/Meeko/blob/07a38f8bac4aa9eef4ef30aa8e67475b15d517aa/meeko/molsetup.py#L1662). 

An additional attribute, `charge_propname`, is added to `MoleculePreparation`, and it needs to be used with option `charge_model = "read"`. The default is `_TriposPartialCharge`. 
The corresponding argument `read_charges_from_prop` is also added to the `from_mol` method for RDKitMoleculeSetup to be accessed at a lower level. It must not be conflicting with argument `compute_gasteiger_charges` (formerly `assign_charges`). 
Some error messages for conflict handling are added. 

Notes: 

- The total charge from the mol2 file will be preserved (i.e. if the total charge is non-integer at the beginning, the sum of the output charges will be the same). 
- In addition to `_TriposPartialCharge`, any other named RDKit Atom attribute can be used as the source of charges. 
- The RDKit parser needs the MOL2 file to have valid SYBYL atom types, which is not the case for some definition MOL2 files in/from Amber/AmberTools. 

Python usage example: 
[212163792.mol2.txt](https://github.com/user-attachments/files/17967004/212163792.mol2.txt)

```py
from meeko import PDBQTWriterLegacy
from rdkit import Chem

input_filename = "212163792.mol2"
output_filename = "212163792.pdbqt"
mol = Chem.MolFromMol2File(input_filename, removeHs=False)
mk_prep = MoleculePreparation(charge_model="read", charge_atom_prop="_TriposPartialCharge")
molsetup_list = mk_prep(mol)
molsetup = molsetup_list[0]
pdbqt_string = PDBQTWriterLegacy.write_string(molsetup)

with open(output_filename, "w") as f:
    f.write(pdbqt_string[0])
```

### DIFF
--- a/meeko/cli/mk_prepare_ligand.py
+++ b/meeko/cli/mk_prepare_ligand.py
@@ -11,6 +11,7 @@ from os import linesep as eol
 import sys
 import json
 import tarfile
+import warnings
 
 from rdkit import Chem
 
@@ -50,12 +51,10 @@ def cmd_lineparser():
     if confargs.config_file is not None:
         with open(confargs.config_file) as f:
             c = json.load(f)
-        
-        for key in c: 
-            if key not in config: 
-                print(f"Error: Got unsupported keyword ({key}) for MoleculePreparation from the config file ({confargs.config_file}). ",
-                      file=sys.stderr,)
-                sys.exit(2)
+        if any(key not in config for key in c): 
+            print(f"Error: Got unsupported keyword ({key}) for MoleculePreparation from the config file ({confargs.config_file}). ",
+                  file=sys.stderr,)
+            sys.exit(2)
         config.update(c)
 
     parser = (
@@ -287,14 +286,14 @@ def cmd_lineparser():
     # check reactive arguments
     if (args.reactive_smarts is None) != (args.reactive_smarts_idx is None):
         print(
-            "Error: Arguments --reactive_smarts and --reactive_smarts_idx require each other",
+            "Arguments --reactive_smarts and --reactive_smarts_idx require each other",
             file=sys.stderr,
         )
         sys.exit(2)
     elif args.reactive_smarts_idx is not None:
         if args.reactive_smarts_idx < 1:
             print(
-                "Error: --reactive_smarts_idx is 1-indexed, but got %d"
+                "--reactive_smarts_idx is 1-indexed, but got %d"
                 % args.reactive_smarts_idx,
                 file=sys.stderr,
             )
@@ -340,14 +339,14 @@ def cmd_lineparser():
         sys.exit(2)
     is_covalent = num_required_covalent_args == 3
     if is_covalent and not _has_prody:
-        msg = "Error: Covalent docking requires Prody which is not installed." + eol
+        msg = "Covalent docking requires Prody which is not installed." + eol
         msg += "Installable from PyPI (pip install prody) or conda-forge (micromamba install prody)"
         print(_prody_import_error, file=sys.stderr)
         print(msg)
         sys.exit(2)
     if min(args.tether_smarts_indices) < 1:
         print(
-            "Error: --tether_smarts_indices is 1-indexed, all values must be greater than zero",
+            "--tether_smarts_indices is 1-indexed, all values must be greater than zero",
             file=sys.stderr,
         )
         sys.exit(2)
@@ -568,7 +567,6 @@ def main():
             config["charge_atom_prop"] = "_TriposPartialCharge"
 
     preparator = MoleculePreparation.from_config(config)
-
     for mol in mol_supplier:
         if is_after_first and output.is_multimol == False:
             print("Processed only the first molecule of multiple molecule input.")
--- a/meeko/cli/mk_prepare_ligand.py
+++ b/meeko/cli/mk_prepare_ligand.py
@@ -11,7 +11,6 @@ from os import linesep as eol
 import sys
 import json
 import tarfile
-import warnings
 
 from rdkit import Chem
 
@@ -33,7 +32,6 @@ else:
 
 
 def cmd_lineparser():
-    backend = "rdkit"
 
     conf_parser = argparse.ArgumentParser(
         description=__doc__,
@@ -48,7 +46,6 @@ def cmd_lineparser():
     confargs, remaining_argv = conf_parser.parse_known_args()
 
     config = MoleculePreparation.get_defaults_dict()
-    print("init config:", config)
 
     if confargs.config_file is not None:
         with open(confargs.config_file) as f:
@@ -56,11 +53,10 @@ def cmd_lineparser():
         
         for key in c: 
             if key not in config: 
-                print(f"Got unsupported keyword ({key}) for MoleculePreparation from the config file ({confargs.config_file}). ")
+                print(f"Got unsupported keyword ({key}) for MoleculePreparation from the config file ({confargs.config_file}). ",
+                      file=sys.stderr,)
                 sys.exit(2)
         config.update(c)
-    
-    print("conf from file:", config)
 
 # region CLIArgParser
     parser = (
@@ -340,7 +336,8 @@ def cmd_lineparser():
     num_required_covalent_args += int(args.tether_smarts is not None)
     if num_required_covalent_args not in [0, 3]:
         print(
-            "Error: --receptor, --rec_residue, and --tether_smarts are all required for covalent docking."
+            "Error: --receptor, --rec_residue, and --tether_smarts are all required for covalent docking.",
+            file=sys.stderr,
         )
         sys.exit(2)
     is_covalent = num_required_covalent_args == 3
@@ -375,7 +372,7 @@ def cmd_lineparser():
         indices[0] = indices[0] - 1  # convert from 1- to 0-index
         indices[1] = indices[1] - 1
 
-    return args, config, backend, is_covalent
+    return args, config, is_covalent
 
 
 class Output:
@@ -506,27 +503,27 @@ class Output:
             return tuple("mk%d" % (i + 1) for i in range(len(molsetups)))
 
 def main():
-    args, config, backend, is_covalent = cmd_lineparser()
+    args, config, is_covalent = cmd_lineparser()
     input_molecule_filename = args.input_molecule_filename
 
     # read input
     input_fname, ext = os.path.splitext(input_molecule_filename)
     ext = ext[1:].lower()
-    if backend == "rdkit":
-        parsers = {
-            "sdf": Chem.SDMolSupplier,
-            "mol2": rdkitutils.Mol2MolSupplier,
-            "mol": Chem.SDMolSupplier,
-        }
-        if not ext in parsers:
-            print(
-                "*ERROR* Format [%s] not in supported formats [%s]"
-                % (ext, "/".join(list(parsers.keys())))
-            )
-            sys.exit(1)
-        mol_supplier = parsers[ext](
-            input_molecule_filename, removeHs=False
-        )  # input must have explicit H
+
+    parsers = {
+        "sdf": Chem.SDMolSupplier,
+        "mol2": rdkitutils.Mol2MolSupplier,
+        "mol": Chem.SDMolSupplier,
+    }
+    if not ext in parsers:
+        print(
+            "*ERROR* Format [%s] not in supported formats [%s]"
+            % (ext, "/".join(list(parsers.keys())))
+        )
+        sys.exit(1)
+    mol_supplier = parsers[ext](
+        input_molecule_filename, removeHs=False
+    )  # input must have explicit H
     
     # configure output writer
     if args.output_pdbqt_filename is None:
@@ -554,7 +551,6 @@ def main():
         rec_prody_mol = prody_parser(rec_filename)
         covalent_builder = CovalentBuilder(rec_prody_mol, args.rec_residue)
 
-    input_mol_counter = 0
     input_mol_skipped = 0
     input_mol_with_failure = (
         0  # if reactive or covalent, each mol can yield multiple PDBQT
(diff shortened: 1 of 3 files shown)


diogomart (commented):
Looks great. Thanks!

Minor requests: 
- If mol2 files are the only source of partial charges in rdkit atom properties (as far as I know they are, but I could be wrong), set default of `charge_propname` to "_TriposPartialCharge" instead of `None` in `MoleculePreparation.__init__` and remove the warning in lines 168-173 of preparation.py. 
- in molsetup.py, keep the warning and default of `read_charges_from_prop` to `None` exactly as they are. Maybe some user is building a molsetup manually and these warnings can be helpful.
- suggestion: rename `charge_propname` in MoleculePreparation to `charge_atom_prop`, and `read_charges_from_prop` also to `charge_atom_prop`. Makes it easier to find connections in the code with `grep`.

diogomart:
Another thing is to add "read" choice to mk_prepare_ligand.py `--charge_model` option.

`raise Warning` triggers an error, we should either raise an error or use `warnings.warn` (needs `import warnings`)

diogomart:
If any other property name was to be used, it would use the MOL/SDF format, as MOL2 charges would always be parsed into the `_TriposPartialCharge` atom property. Looks like SDF can store atom properties:
https://github.com/openforcefield/openff-toolkit/issues/250
https://github.com/rdkit/rdkit/pull/2297
so I guess we could expose also `--charge_atom_prop` in mk_prepare_ligand.py, and even go to the extent of having mk_prepare_ligand.py set `charge_atom_prop="PartialCharge"` if the input is SDF, and `charge_atom_prop="_TriposPartialCharge"` if MOL2. I'm not sure anymore about setting the default `charge_atom_prop` to "_TriposPartialCharge".

rwxayheee:
Hi @diogomart 
Thanks for looking at this PR. 

- About the default property name

The `None` was to explicitly handle possible conflicts of `charge_model`/`compute_gasteiger_charges`(formerly `assign_charge`): when a valid property name is given, all other built-in charge models will be considered conflicting. I kind of want to keep the default to be None, because when `charge_model` is `read`, a default property name becomes "_TriposPartialCharge". https://github.com/forlilab/Meeko/pull/258/files#diff-8fe710b00c5629fbd7adc003157cd2d4a1e204a15eb4fb0f38221f7970300712R167-R169 is where that happens. 
Therefore, the default property name is "_TriposPartialCharge" when there isn't a conflicting charge model. 

- About the two classes that both use the property name keyword arguments

`MoleculePreparation` is on a higher level than `RDKitMoleculeSetup` (keyword arguments flow from `MoleculePreparation` to `RDKitMoleculeSetup`). Both of them can be used on a molecule with manually setup charges (?). So I set up the default values for both in the same way. But maybe that's redundant.. I copied and pasted most of the argument handling part. 

- About the different keyword argument names 

I did this on purpose because they have different dominance over other options. `read_charges_from_prop` means to read AND to read from a specified property. Whereas `charge_propname` only specifies the property name, but whether to read depends on `charge_model` (and this is also for conflict handling, but maybe this is all unnecessary..). I can change the names to make tracking easier. 

- About SDF support
The PR is to support charges from MOL2, but I will try reading charges from SDF. There are other sources to read charges from (Amber, QChem files) but as long as users can store them into a custom atom property, we just need the property name. 

Tomorrow I will work on: 
- Keyword name changes
- Warning and CLI options
- Parsing charges from SDF, usually it doesn't store partial charges (only formal charges) in the molecule section, but the charges could be parsed from the property field (??? I don't know)

diogomart:
Ok, all of that sounds good to me. SDF is already supported, it's just a different atom property name.

rwxayheee:
> Ok, all of that sounds good to me. SDF is already supported, it's just a different atom property name.

(If you happen to have one) could you provide an SDF with partial charges? Do people write partial charges in property field or somewhere else? The SDF on PubChem usually doesn’t come with partial charges. But other software / packages might be able to write them in custom fields. The charges will be parsed as molecule property but not atom property. This PR currently needs charges in atom property 

no worry though, I can look that up from some other databases tomorrow.

diogomart:
They can actually be parsed to an atom property (I learned that today), see my post above, the link to an rdkit PR 2297 has an example SDF with charges.

rwxayheee:
Ahhhh!!! I see, thanks! I will look into that

rwxayheee:
Summary of added commits: 

- Changed default atom property name to `PartialCharge` 

- Changed keyword and variable name `charge_propname` to `charge_atom_prop`

- Use `warnings.warn` instead of `raise Warning` to not interrupts the flow of execution

- Added argument `--charge_from_prop` to `mk_prepare_ligand.py`
Unlike how other arguments are passed, their changes are made in the `main` function, and right before the config right before `MoleculePreparation.from_config` because we need access `ext` (extension of input files) to determine the right property name. 

Expected CLI usage: 

```
# switch property name based on extension
mk_prepare_ligand.py -i 212163792.mol2 --charge_model read 

# customize property name 
mk_prepare_ligand.py -i 212163792.mol2 --charge_model read --charge_from_prop "_TriposPartialCharge"
```

The following will not work: 

```
mk_prepare_ligand.py -i 212163792.mol2 --charge_from_prop "_TriposPartialCharge"
```
Because the default charge_model in `mk_prepare_ligand.py` is "gasteiger". 

I tested with a few Python and command-line script examples. @diogomart thanks again for your suggestions, and please let me know what you think!

meeko/cli/mk_prepare_ligand.py
```diff
@@ -78,6 +78,11 @@ def cmd_lineparser():
         "--name_from_prop",
         help="set molecule name from RDKit/SDF property",
     )
+    io_group.add_argument(
+        "--charge_from_prop",
+        nargs="*",
```
diogomart: no need for nargs

meeko/cli/mk_prepare_ligand.py
```diff
@@ -551,6 +556,19 @@ def main():
...
     preparator = MoleculePreparation.from_config(config)
+
+    if args.charge_from_prop is not None: 
+        if args.charge_model != "read": 
+            print('--charge_from_prop must be used with --charge_model "read"')
+            sys.exit(1)
+
+        preparator.charge_model = "read"
+        if args.charge_from_prop: 
+            preparator.charge_atom_prop = args.charge_from_prop[0]
```
diogomart: no need for `[0]` after nargs above is removed

meeko/cli/mk_prepare_ligand.py
```diff
@@ -551,6 +556,19 @@ def main():
...
+        else:
+            if ext=="mol2": 
+                preparator.charge_atom_prop = "_TriposPartialCharge"
```
diogomart: this is inside the `if args.charge_from_prop is not None` so i think this never gets executed

rwxayheee:
Hi @diogomart 
Thanks for looking at this. Sorry for not getting this right. I missed an "if" statement after I rewrote something else. But I noticed another problem. 

```py
    if args.charge_from_prop is not None: 
        if args.charge_model != "read": 
            print(f'--charge_from_prop must be used with --charge_model "read", but the current charge_model is "{args.charge_model}". ')
            sys.exit(1)
        config["charge_atom_prop"] = args.charge_from_prop if args.charge_from_prop else "_TriposPartialCharge" if ext == "mol2" else None
    else:
        if args.charge_model == "read": 
            if ext=="mol2": 
                config["charge_atom_prop"] = "_TriposPartialCharge"
```

Here I wanted to override config["charge_atom_prop"] according to arguments and `ext` (extension of input), which is only accessible in the `main` function without further changes. But it basically ignores the values from the original config and it's not right to do that. I will fix it a little later today. Sorry for my confusion!

rwxayheee:
0568941 should fix the current issues that are specific to the added arguments. 

There might still be a small problem with how command line arguments override config, specifically those with a default value for all other command-line arguments, like `charge_model`. These default default values always override the config file (?), but the overriding should happen only when the arguments are explicitly provided. 

Something like the following might work: 

```py
    # command line arguments override config
    explicitly_provided_args = {
        arg.lstrip('-').replace('-', '_') for arg in remaining_argv if arg.startswith('--')
    }

    for key in config:
        if key in explicitly_provided_args:
            config[key] = args.__dict__[key]
```

To handle arguments that are provided in the short-form, a mapping can be created, like: 

```py
    # Mapping arguments short-form to long-form 
    short_to_long = {}

    for action in parser._actions:
        option_strings = action.option_strings
        
        short_option = next((opt for opt in option_strings if opt.startswith('-') and not opt.startswith('--')), None)
        long_option = next((opt for opt in option_strings if opt.startswith('--')), None)
        
        if short_option and long_option:
            short_to_long[short_option.lstrip('-')] = long_option.lstrip('--')

    explicitly_provided_args = explicitly_provided_args | {
        short_to_long[arg.strip('-')] for arg in remaining_argv if arg.startswith('-') and not arg.startswith('--')
    }
```

But maybe it's all unnecessary. @diogomart Could you please confirm that we want to keep the config option, and should make the changes to only allow argument overriding config for arguments that are explicitly provided? Does the current type of overriding align with what we want, and it's not necessary to change? 

I will come back and take a closer look later this evening. Thank you for your time and kind advice in advance.

rwxayheee:
The charge parsing should be working now, hopefully.. Also fixed some small problems with command line arguments defaults unexpectedly overriding config options. The changes are, specifically: 

- When reading options from config file, check if the keyword is accepted by MoleculePreparation. (Not necessaary; removed in [81151af](https://github.com/forlilab/Meeko/pull/258/commits/81151af6e7b26ada854318527930ec3c2837a2a1))

- Removed one redundant assignment from args to config

```py
config["load_atom_params"] = args.load_atom_params
```

Because config should already be synced after 

```py
config[key] = args.__dict__[key]
```

- `args.add_atom_types_json` and config["add_atom_types"] are handled differently (from other args) because they have different names. The script used to ignore the existing assignments in config

- config["charge_atom_prop"] is handled differently because it might depend on file extension, which function `cmd_lineparser` doesn't have access to. 

- Cleaned up the unused `backend = "rdkit"` blocks; redirect error messages to Standard Error

meeko/cli/mk_prepare_ligand.py
```diff
@@ -52,7 +51,11 @@ def cmd_lineparser():
     if confargs.config_file is not None:
         with open(confargs.config_file) as f:
             c = json.load(f)
-            config.update(c)
+        if any(key not in config for key in c): 
+            print(f"Error: Got unsupported keyword ({key}) for MoleculePreparation from the config file ({confargs.config_file}). ",
```
diogomart: key is not defined in the f string (that's why we iterate over the keys in MoleculePreparation.from_config)

diogomart: i'm not sure we need to test the keys here, `MoleculePreparation.from_config` does a check

meeko/cli/mk_prepare_ligand.py
```diff
@@ -78,6 +81,10 @@ def cmd_lineparser():
         "--name_from_prop",
         help="set molecule name from RDKit/SDF property",
     )
+    io_group.add_argument(
+        "--charge_atom_prop",
+        help="set atom partial charges from an RDKit atom property based on the input file. The default is 'PartialCharge' for SDF and '_TriposPartialCharge' for MOL2 unless overriden by a user defined property name. ",
```
diogomart: I would move this next to `--charge_model` in the Molecule Preparation group. While it is certainly related to I/O, being next to charge_model is more important, and charge_model belongs in molecule preparation.

meeko/cli/mk_prepare_ligand.py
```diff
@@ -293,13 +300,10 @@ def cmd_lineparser():
             sys.exit(2)
         args.reactive_smarts_idx -= 1  # convert from 1- to 0-index
 
-    # command line arguments override config
-    for key in config:
-        if key in args.__dict__:
+    for key in args.__dict__:
```
diogomart: i would like the "command line arguments override config" comment back :)

diogomart (commented):
This is great. Thank you for all the clean up. Great catch with the add_atom_types option.

We do want to keep the option to pass a JSON config file, and have any command line arguments override those in the config file, but only if they are explicitly passed. This is what is implemented, and based on some testing it is working as expected.

I have only minor requests, which are in-lined. Thanks again!

meeko/cli/mk_prepare_ligand.py
rwxayheee: You're right on both. I made the unnecessary changes but I forgot to revert.

meeko/cli/mk_prepare_ligand.py
rwxayheee: This is unnecessary, I would revert that

meeko/cli/mk_prepare_ligand.py
rwxayheee: ok

rwxayheee:
Hi @diogomart thanks again for the comments. Sorry this took so many commits; I hope things are ok now. I did not understand the overriding at first, so I did too much rewrite that was unnecessary.

diogomart:
Thanks for the edits. The overriding is done in a complicated way... couldn't come up with a better one. I just pushed a commit to add a couple comments and moved a block of code to make it more understandable. Sorry for the lack of comments!

diogomart (approved):

rwxayheee:
Hi @diogomart Thanks so much for the comments and for your kind support, as always, merging!

PR #259, opened 2024-12-01
Improve basic support for elements (single-atom residues, metal, noble gas) that are unsupported by ad4

### LINKED ISSUE #252: Support for metal atoms
I'm using the cli to prepare metalloproteins for docking, however, it seems to have issues with some metals:

`RuntimeError: Element Na doesn't have an implemented covalent radius, which was required for the perception of intermolecular bonds` I encounter this error for a couple elements like Na, Co, Ni, Pt, Yb and K.

It seems meeko uses the [rdkit periodic table module](https://www.rdkit.org/docs/source/rdkit.Chem.PeriodicTable.html) to get the radii. Is there a way to bypass it or use a dummy value? Or maybe it could be added to be considered under the `--allow_bad_res` flag.

rwxayheee:
This is for #252. Although still very primitive, it has the basic support to allow all elements (1-118 *is this even necessary*), single-atom residues, metal or noble gas to be in receptor (polymer). 

Here's a summary of major changes: 

1. Removed redundant reference tables: `covalent_radius` from `meeko.utils.rdkitutils`, `mini_periodic_table` from `meeko.utils.utils`
They were replaced by the equivalent references, which are: 
```py
from .utils.covalent_radius_table import covalent_radius
from meeko.utils.autodock4_atom_types_elements import autodock4_atom_types_elements
periodic_table = Chem.GetPeriodicTable()
```
Made some other minimal adjustments in accordance to the replacement. 

2. When constructing a polymer, always not make intermol (between two monomers) bonds involving metal or noble gas, since there isn't a good solution for padding either. 

3. Allow single-atom residues and stop checking for unsupported elements in template generation. 
Single-atom residues like Co2+, Ar, etc., although not registered, they can now be fetched from PDB (CCD) and will have integer charges unless overridden by user. The template will always be built following the unbound ligand routine not linked. 

4. For receptor preparation involving ad4-unsupported elements, it's always possible to generate the receptor json/pdb file but the PDBQT file requires a valid atom type in the parameter file. Unspecified atom type shall not be forgiven and the error message will be: 

```
RuntimeError: Cannot write PDBQT file. The following atom does not have an atom type: 
ATOM   3411  CO  CO  A 501       4.245  -2.575  79.904  1.00  0.00     2.000 None
Consider declaring a custom atom type in the parameter file or revising the input structure. 
```

5. Some small fixex for CLI script `mk_prepare_receptor.py`
- Print error message and exit, when ProDy parser is invoked but extension of input file is unsupported. 
- Fixed a small problem with `--write_pdb` 
- polymer.to_pdb() writes element in columns 78-79, which should be 77-78. This made our output PDB not readable by some parsers (including mk_prepare_receptor iteself). Fixed in [9391c01](https://github.com/forlilab/Meeko/pull/259/commits/9391c01f02859e5a9aca19fab21bc84b165dac41)


rwxayheee:
To do: 

- [ ] Resolve conflicts
- [ ] Add error handling and skip espaloma graph construction for single-atom residues in cases like #310 
- [ ] Review supported elements by espaloma (openff toolkit), add warning as needed

rwxayheee:
Closing due to excessive merge conflicts. A new PR will be opened from a fresh branch.

Issue #260, opened 2024-12-01
RecursionError: maximum recursion depth exceeded while getting the repr of an object


ivnsch:
Hi, I just installed meeko from conda-forge and am following the basic docking for vina tutorial https://meeko.readthedocs.io/en/release/tutorial1.html#for-autodock-vina

```
pdb_file="Meeko/example/tutorial1/input_files/1iep_protein.pdb"
mk_prepare_receptor.py --read_pdb $pdb_file -o rec_1iep -p
```

and I get
```
  File "<path>/miniconda3/envs/vina/bin/mk_prepare_receptor.py", line 7, in <module>
    from meeko.cli.mk_prepare_receptor import main
  File "<path>/miniconda3/envs/vina/lib/python3.11/site-packages/meeko/__init__.py", line 10, in <module>
# ... 
RecursionError: maximum recursion depth exceeded while getting the repr of an object
```

What's wrong? thanks!

rwxayheee:
Can confirm this problem may be a result of circular import. When installed via conda, the generated script is

```py
#!/bin/sh
'''exec' /opt/homebrew/anaconda3/envs/meeko_tutorial/bin/python "$0" "$@"
' '''
# -*- coding: utf-8 -*-
import re
import sys
from meeko.cli.mk_prepare_receptor import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

This problem is not observed with some other package managers like micromamba. As a walkaround for the time being, you could install meeko via pip or micromamba. 

We should fix this; we might want to rename the console scripts, or devert the entry points.

rwxayheee:
Hi @ivnsch 
Thanks again for reporting this. I opened a PR (#261) but need to talk to other lab members about it. We will need to upload the updated package and it might take a little longer. While we're working on this, there are a few walkaround options: 
- `micromamba install` the current package; micromamba constructs the console scripts differently so we dodged the problem
- `pip install` from this repository
-  (last but not least) `conda install` from a built package with #261. I validated the change this way

diogomart:
I can't reproduce on linux. @rwxayheee did you observe the recursion error?

rwxayheee:
Hi @diogomart! Yes, I can reproduce this when installed via conda on Mac.

diogomart:
Do you know where the circular import happens?

rwxayheee:
[...]

PR #261, opened 2024-12-01
Use dynamic entry to avoid circular import in generated console scripts when installed via conda

rwxayheee:
This is for #260, it seems to work, but might not be the best solution..


rwxayheee:
This does not fix the importlib issues that were potentially caused by outdated entry point, or files from a different version that were not properly renamed/updated in version change.

PR #266, opened 2024-12-07
Format string in output file with plain line-ending character instead of os.linesep

rwxayheee:
The issue only affects Windows PowerShell. In some files written by Meeko, each line has double the line-ending characters. This is likely because the Python built-in function `f.open()` by default translates '\r\n' (os.linesep on Windows) to '\r\r\n' when writing the file, which will be read as '\n\n' when the file is re-opened. 

https://docs.python.org/3/library/functions.html#open

> When writing output to the stream, if newline is None, any '\n' characters written are translated to the system default line separator, [os.linesep](https://docs.python.org/3/library/os.html#os.linesep). If newline is '' or '\n', no translation takes place. If newline is any of the other legal values, any '\n' characters written are translated to the given string. 

This PR replaces all eol by the plain line-ending character '\n'. Another possible solution is to always write without translation, i.e. f.open(newline='') but that isn't the default. Thanks to Nicole Benson (Fuchs lab, OSU) for bringing this to our attention.

### DIFF
--- a/meeko/cli/mk_export.py
+++ b/meeko/cli/mk_export.py
@@ -7,7 +7,7 @@ import argparse
 import gzip
 import pathlib
 import sys
-from os import linesep as eol
+eol="\n"
 import warnings
 import copy
 import numpy as np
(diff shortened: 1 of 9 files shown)


diogomart (approved):
Thanks :-)

rwxayheee:
Thanks! Merging to develop

PR #270, opened 2024-12-08
don't max out substructure matches at 1000

diogomart:
Unreasonably large ligands will have untyped atoms without this.


rwxayheee (approved):
This fixed the reported issue. thanks!

Issue #271, opened 2024-12-09
Use try/except in mk_prepare_ligand.py to automatically exclude "bad" molecules


diogomart:

rwxayheee:
Fixed in #272. I implemented a very basic try/except to skip mols with problems in prepare (covalent docking, which currently uses a different routine, was not covered).

PR #272, opened 2024-12-09
Catch prepare errors without interrupting batch processing

rwxayheee:
This is a quick fix to catch prepare errors without interrupting batch processing, using try/except as proposed in #271.


diogomart (approved):
Thanks!

PR #273, opened 2024-12-09
Complicated metal complexes as ligand

### LINKED ISSUE #267: RuntimeError: expected only 1 added H per heavy atom, maybe deleted element had double bond to this heavy atom
Hi, I'm trying to prepare ligands for virtual screening. I have a multi-molecule SDF file of ligands prepared using RDKit which I need to split into multiple PDBQT files to serve as input for QuickVina2-GPU-2.1. 

I tried running this:

`mk_prepare_ligand.py -i compounds.sdf --rigid_macrocycles --multimol_outdir .
`

It runs for a while and then throws this error:

> RuntimeError: expected only 1 added H per heavy atom, maybe deleted element had double bond to this heavy atom

I'm guessing there's something wrong with some compound which I'm unable to figure out. I have attached the multi-molecule SDF file for reference. Please help.
Thanks.
[compounds.zip](https://github.com/user-attachments/files/18047922/compounds.zip)

rwxayheee:
This small edit on the charge method allows potentially complicated metal complexes (with datives, double, triple or quadruple bonds) to become ligand, where metal gets a partial charge as the sum of the metal's formal charge (from input) and the Gasteiger charges of non-real Hs that were added to compensate breaking M-L bonds. 

It should also support charge assignment in bridged structures like M-L-M. Here, the coordinate atom L will have multiple added Hs, and the charge will be distribute to both metal atoms based on the bond order. 

Dative bonds, are a special case. Breaking a dative bonds do not result in addition of non-real Hs. 

This process looks complicated. I put some more in-line comments, but in future we could rewrite it.


meeko/molsetup.py
```diff
@@ -1639,27 +1639,76 @@ def remove_elements(mol, to_rm=(12, 20, 25, 26, 30)):
...
+                    atoms_in_mol[nei_idx].SetNumExplicitHs(nei_valence + num_nr_h)
+                    
         rwmol = Chem.EditableMol(mol)
         for idx in sorted(idx_to_rm, reverse=True):
             rwmol.RemoveAtom(idx)
         mol = rwmol.GetMol()
-        for idx in neigh_idx_to_nr_h:
-            n = neigh_idx_to_nr_h[idx]
-            newidx = idx - sum([i < idx for i in idx_to_rm]) 
-            mol.GetAtomWithIdx(newidx).SetNumExplicitHs(n + 1)
```
rwxayheee, line 1689: SetNumExplicitHs without Resetting it (since the added hydrogens are non-real) might be causing some problems when functions like RemoveHs are called later on the ligand mol

rwxayheee: Instead of resetting it, not modifying atoms in input mol is better

rwxayheee:
This PR allows preparation of metal complexes with dative bonds as ligands. An example is heme (with Fe(II): 
Chemical structure:
heme_mine_structure

Input SDF: 
[heme_mine.sdf.txt](https://github.com/user-attachments/files/18072521/heme_mine.sdf.txt)

Output PDBQT: 
[heme_mine.pdbqt.txt](https://github.com/user-attachments/files/18072528/heme_mine.pdbqt.txt)

rwxayheee:
Merged develop into this branch. 

Added a test (example showed in this PR): 
- added a script file `metal_test.py`
- added a folder `metal_complex_data` for test files

meeko/molsetup.py
```diff
@@ -1665,31 +1665,80 @@ def from_mol(
...
-            if atom.GetAtomicNum() in to_rm:
-                idx_to_rm[atom.GetIdx()] = atom.GetFormalCharge()
-                rm_to_neigh[atom.GetIdx()] = set()
-                for neigh in atom.GetNeighbors():
-                    n = neigh.GetNumExplicitHs()
-                    neigh_idx_to_nr_h[neigh.GetIdx()] = n
-                    rm_to_neigh[atom.GetIdx()].add(neigh.GetIdx())
         if not idx_to_rm:
-            return Chem.Mol(mol), idx_to_rm, rm_to_neigh
+            return input_mol, idx_to_rm, rm_to_neigh
```
diogomart: Need to either return a copy of input_mol, or make a copy before changing Se to S for computing Gasteiger charges in the function that calls `remove_elements()`

diogomart (changes requested):

meeko/molsetup.py
rwxayheee: Can you explain why and what's the difference?

diogomart: It's to prevent a side effect (a modification) on a molecule that a user passes to an instance of MoleculePreparation. This could be a padded_mol in a Polymer. The modification is changing Selenium to Sulfur here: https://github.com/rwxayheee/Meeko/blob/cb7b4e7db4334f161b35124135214ca69099c3f1/meeko/molsetup.py#L1768-L1771
That's why the original code returns a copy of the input molecule, instead of returning the input molecule itself. Then the copy is modified, and the original input mol is safe from side effects.

rwxayheee: Got it, thanks, I understand now

Issue #269, opened 2024-12-08
Issue: Maximum Recursion Depth Exceeded in mk_prepare_ligand.py


hezhizhang-269:
Hello,

When I was running mk_prepare_ligand.py to process a large .sdf file, the script encounters a RecursionError, indicating that the maximum recursion depth has been exceeded. This error appears to occur during the molecule's ring perception process, preventing successful preparation of the ligand.

The input .sdf file is attached here [3v7e_model1_aptamer_Model_1_with_heterogeneous_atoms_removedH_no_valence_issue_3.sdf.txt](https://github.com/user-attachments/files/18053577/3v7e_model1_aptamer_Model_1_with_heterogeneous_atoms_removedH_no_valence_issue_3.sdf.txt).

The command I used to run the script and the outputs were as follows:

```
caomq@cscigpu06:~/3v7e$ mk_prepare_ligand.py -i 3v7e_model1_aptamer_Model_1_with_heterogeneous_atoms_removedH_no_valence_issue_3.sdf -o 3v7e_model1_aptamer_Model_1_with_heterogeneous_atoms_removedH_no_valence_issue_3.pdbqt
/home/caomq/.local/lib/python3.8/site-packages/scipy/fft/__init__.py:97: DeprecationWarning: The module numpy.dual is deprecated.  Instead of using dual, use the functions directly from numpy or scipy.
  from numpy.dual import register_func
/home/caomq/.local/lib/python3.8/site-packages/scipy/special/orthogonal.py:81: DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`. To silence this warning, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
  from numpy import (exp, inf, pi, sqrt, floor, sin, cos, around, int,
/home/caomq/.local/lib/python3.8/site-packages/pandas/core/computation/expressions.py:20: UserWarning: Pandas requires version '2.7.3' or newer of 'numexpr' (version '2.7.1' currently installed).
  from pandas.core.computation.check import NUMEXPR_INSTALLED
/usr/lib/python3/dist-packages/apport/report.py:13: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
[...]

PR #276, opened 2024-12-10
Remove from class Ring the unused attributes to reduce overhead in molecule preparation

### LINKED ISSUE #275: Reduce overhead in preperation
@rwxayheee we might be able to remove those lines of code. They compute a graph that is stored in instances of `Ring`. I think those graphs are never used, along with other attributes that are probably also never used, see https://github.com/forlilab/Meeko/pull/109#discussion_r1588546970

_Originally posted by @diogomart in https://github.com/forlilab/Meeko/issues/269#issuecomment-2527317732_
            

Currently looking at molsetup.py to see if we could reduce overhead during molecule preparation, or remove unused functions to make it more light-weighted

rwxayheee:
Remove from class Ring the unused attributes: graph and corner_flip

### DIFF
--- a/meeko/macrocycle.py
+++ b/meeko/macrocycle.py
@@ -83,9 +83,7 @@ class FlexMacrocycle:
             setup.rings.keys()
         ):  # ring_id are the atom indices in each ring
             size = len(ring_id)
-            if setup.rings[ring_id].is_aromatic:
-                rigid_rings.append(ring_id)
-            elif size < self._min_ring_size:
+            if size < self._min_ring_size:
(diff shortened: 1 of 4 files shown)


rwxayheee:
Hi @diogomart 
Thanks for adding the test! Could we merge this, or shall we hold for a bigger refactoring of rings perception? 
This effectively avoids the original issue in #269

diogomart (approved):
It's good to merge!

rwxayheee:
Thanks!! Merging this to develop

PR #287, opened 2024-12-16
--cache_templates for mk_prepare_receptor.py

### LINKED ISSUE #209: Cache templates downloaded from CCD

rwxayheee:
This is an alternate to #209. Instead of caching CIF files (which are not really useful for other functions in Meeko), the encoded residue chem templates will be cached.

### DIFF
--- a/meeko/cli/mk_prepare_receptor.py
+++ b/meeko/cli/mk_prepare_receptor.py
@@ -179,8 +179,10 @@ def get_args():
                                   "(optionally) a provided JSON filename. " 
                                   "Default cache filename is: $HOME/.meeko_residue_chem_templates_cached.json) "
                                   "When the caching is ON, the templates for polymer construction will be read from "
-                                  "the specified cache file and updates may be made to the same file in a cumulative manner. "
-                              )
+                                  "the specified cache file and updates may be made to the same file in a cumulative manner. ", 
+                              ), 
+                              nargs = "?", 
+                              default = False, 
     )
     config_group.add_argument("--mk_config", help="[.json]", metavar="JSON_FILENAME")
     config_group.add_argument(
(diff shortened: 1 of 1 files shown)


rwxayheee:
New argument added: 

```
  --cache_templates [CACHE_TEMPLATES]
                        Turns on caching of ResidueChemTemplates (default is OFF) by this option and (optionally) a
                        provided JSON filename. Default cache filename is: $HOME/.meeko_residue_chem_templates_cached.json)
                        When the caching is ON, the templates for polymer construction will be read from the specified
                        cache file and updates may be made to the same file in a cumulative manner.
```

Usage note: 

(1) Creates, updates cumulatively and uses default cached templates

```
mk_prepare_receptor.py -i 6O4W.pdb -o some_output --cache_templates
```

This usage creates and updates the cached templates at `$HOME/.meeko_residue_chem_templates_cached.json`, with a message: 
`--cache_templates is turned on, but a name is not provided. The default filename ($HOME/.meeko_residue_chem_templates_cached.json) will be used.` 

(2) Specifies a destination or re-uses an existing cache

```
mk_prepare_receptor.py -i 6O4W.pdb -o some_output --cache_templates hohoho.json
```

This usage creates and updates the cached templates at the given name.

diogomart:
In my use cases, the most common cause for downloading a template repeatedly from the PDB is when running mk_prepare_receptor.py multiple times to identify and fix issues (bad residues, alt locs, etc). What do you think of saving the template by default in the current working dir?

diogomart:
and then, before downloading it, first check if the template exists in the working dir. Templates could be prefixed with "downloaded_CCD_" or similar and added to the list of files that mk_prepare_receptor.py writes that is printed for each run of mk_prepare_receptor.py

rwxayheee:
> In my use cases, the most common cause for downloading a template repeatedly from the PDB is when running mk_prepare_receptor.py multiple times to identify and fix issues (bad residues, alt locs, etc). What do you think of saving the template by default in the current working dir?

`args.cache_templates` (`--cache_templates <path to cached template file`>) is customizable. The current default path is in the home directory so the templates are saved to one single place, regardless of the current work directory. 

Would you like to change the default?

diogomart (approved):

Issue #288, opened 2024-12-17
fix from_json methods, add from_json_file and from_dict methods


diogomart:
Some `from_json` methods in take Python dictionaries as argument. These methods should be renamed to `from_dict`. Then, `from_json` methods wrap the `from_dict` ones. And `from_json_file` wrap the `from_json` ones.

rwxayheee:
In molsetup.py, some are decoders (like the `MoleculeSetup` json decoder) named as `from_json`. It can be named `molecule_setup_decoder`, to be consistent with what's in polymer.py. 

@diogomart To make the method names consistent, would you be interested in the idea of making these basic methods (to_json, to_json_file, from_json, from_json_file) inheritable from a master (base) class? Earlier, I wanted to do the same thing in another issue
https://github.com/forlilab/Meeko/issues/279#issuecomment-2540143906

I was also wondering if you might be interested in moving encoders and decoders into their own respective class? The expected keys, and the encode and decode function will be defined within the scope of the respective class. When there's a layered structure, then higher-level class can reference the encoders/decoders in the lower-level class. I'm not sure if this is common or if there's any efficiency drawback, but it might make maintenance easier. Let me know what you think, and I'm willing to take the work!

diogomart:
Inheriting for these JSON interface methods carries the risk of having exceptional situations later and being unable to adhere to the base class. Maybe we'll have some new object for which reading from JSON doesn't make sense, then we'll choose to not inherit from it partially defeating the purpose of a base class, or have to find hacky ways to inherit anyway but not implement all the functions. At first sight I rather not inherit, but I'll think about it more... I like the features of the base class in #279

We also can't return the object passed as input (a dictionary) if the expected keys don't match: we need to raise an error. For example, if we dump a Polymer to JSON with v0.6.1 and load it again with a more recent version after PR #276 we get dictionaries instead of MoleculeSetup objects because the ring_corners attribute is not expected anymore. Raising an error would have allowed use to catch this in the tests. We need to create an exception like this one from #254 https://github.com/forlilab/Meeko/blob/8780dd052acc7a26e3c2c10aa2cc0ac9615e0155/meeko/molsetup.py#L332-L342 so that Polymer JSONs written with v0.6.1 will be supported in later versions.

rwxayheee:
@diogomart thanks so much for your response. All this makes sense to me, and I understand it's difficult to implement at the moment. 

[...]

Issue #279, opened 2024-12-11
Catch and elaborate errors for corrupt json or incorrect object hook


rwxayheee:
There are several types of JSON files in the Meeko world. Many objects can be created from JSON (which is really convenient when put into use), but the errors are not so clear if there's a problem within the JSON file or the incorrect object hook. It might be helpful to catch and elaborate errors (just provide some quick hints) for bad input JSON or incorrect object hook. 

There are some situations that the JSON can be loaded into a dict, but cannot be built into a valid Python object like Polymer. I wish there's a way to catch the problem, but I'm not so sure how to do it

rwxayheee:
Its really really not necessary unless more people want it. Here's a class to inherit from. Something like the following gives some hints on where the problem occurred (parsing, decoder function, object type) and avoids letting go a partial parse silently. 

```py
class BaseJSONParsable:

    # Define in Individual Subclasses 
    expected_json_keys = None
    object_hook = None 
    # and some decoder function

    # Inherit from_json and from_json_file
    @classmethod
    def from_json(cls, json_string):
        try: 
            obj = json.loads(json_string, object_hook=cls.object_hook)
            # Log mismatched keys (maybe not a fatal problem)
            if set(obj.keys()) != cls.expected_json_keys:
                logging.error(
                    f"Keys from JSON ({set(obj.keys())}) differ from "
                    f"expected keys ({cls.expected_json_keys})."
                )  
            # Success
            if isinstance(obj, cls): 
                return obj

        except Exception as decoder_error:     
            try: 
                obj = json.loads(json_string)
            # Error occurred while parsing JSON
            except Exception as parser_error: 
                raise RuntimeError(
                    f"Unable to parse the source JSON for {cls.__name__}: {parser_error}"
                )
            # Error occurred within the decoder function
            raise RuntimeError(
                f"An error occurred when creating {cls.__name__}: {decoder_error}"
            )
        # Failed to create obj
        raise ValueError(
            f"Unexpected object type created from JSON: {type(obj)}. "
            f"Expected object type is: {cls.__name__}."
        )
            
    @classmethod
    def from_json_file(cls, json_file): 
        with open(json_file, "r") as f: 
            json_string = f.read()
        return cls.from_json(json_string)
```

PR #292, opened 2024-12-18
Introduce a base class for classes with JSON-interchange functions

rwxayheee:
This PR introduces a base class for many other existing classes that need the JSON-interchange functionalities. The common, highe-level functions (from_json, to_json) are inherited from the base class. Whereas the specific encoding-decoding logics are defined within the individual subclass scope. 
The major objective is to reduce code duplication, improve error handling, and make future maintenance easier. This may or may not be an over-engineering of simple things that are already working, but I thought it would be really useful when I opened #279. I hope this can also help with the consistency issue in #288. 

- [x] Introduce `BaseJSONParsable` at 5ec1bf7
This working checkpoint shows the basic layout of class `BaseJSONParsable`. Note that `RDKitMoleculeSetup` is still a subclass of `MoleculeSetup`. For subclasses with additional inheritance relationships like this, the decoder/encoder of the parent may be used in the child class. 

- [x] ResidueChemTemplates: Harmonize from_json and from (data) json at 1cd7a77 
This working checkpoint shows how the parameter data JSON string and the standard serialized JSON of ResidueChemTemplates may be parsed in the same way. 

- [x] MoleculePreparation: harmonize from_config and from (deserialized json object) dict at 252bc95
For https://github.com/forlilab/Meeko/issues/288#issuecomment-2552078213

- [x] Revisit optional_keys
Future changes like #254 should be possible by changes made to __init__ and the JSON-interchange region in the subclass.


rwxayheee:
This is done, but open to any changes it may need. Here's a very brief summary: 

### Implemented code structure in this PR

- A subclass of `BaseJSONParsable` needs the following assets that contain subclass-specific details: 

```py
    expected_json_keys: Optional[frozenset[str]] = None

    @classmethod
    def json_encoder(cls, obj):
        raise NotImplementedError("Subclasses must implement json_encoder.")
    
    @classmethod
    def _decode_object(cls, obj):
        raise NotImplementedError("Subclasses must implement _decode_object.")
```

- `BaseJSONParsable` provides the other high-level functions that are sharable: 

```py
    @classmethod
    def json_decoder(cls, obj: dict[str, Any], check_keys: bool = True):
        # Avoid using json_decoder as object_hook for nested objects
        if not isinstance(obj, dict):
            return obj
        if check_keys and not cls.expected_json_keys.issubset(obj.keys()):
            return obj

        # Delegate specific decoding logic to a subclass-defined method
        return cls._decode_object(obj)

    @classmethod
    def from_json(cls, json_string: str):
        try: 
            # Log mismatched keys if non-critical 
            obj_dict = json.loads(json_string)
            actual_keys = obj_dict.keys()
            if actual_keys != cls.expected_json_keys:
                missing_keys = cls.expected_json_keys - actual_keys
                extra_keys = actual_keys - cls.expected_json_keys
                logging.warning(
                    f"Key mismatch for {cls.__name__}. "
                    f"Missing keys: {missing_keys}, Extra keys: {extra_keys}."
                )

            try: 
                obj = json.loads(json_string, object_hook=cls.json_decoder)
            
            # Error occurred within cls.json_decoder
            except ValueError as decoder_error: 
                raise RuntimeError(
                    f"An error occurred when creating {cls.__name__} from JSON."
                    f"Error: {decoder_error}"
                )

            # Validate the decoded object type
            if not isinstance(obj, cls):
                raise ValueError(
                    f"Decoded object is not an instance of {cls.__name__}. "
                    f"Got: {type(obj)}"
                )
            return obj

        # Error occurred within json.loads
        except Exception as parser_error: 
            raise RuntimeError(
                    f"Unable to load JSON string for {cls.__name__}. "
                    f"Error: {parser_error}"
            )

    @classmethod
    def from_dict(cls, obj: dict) -> "BaseJSONParsable":
        return cls.json_decoder(obj, check_keys=False)
    
    def to_json(self):
        return json.dumps(self, default=self.__class__.json_encoder)
```

Function `_decode_object` focuses on the decoding logic, while `from_json` is on a higher level and is more for error handling. On the intermediate level, there's `json_decoder` as `object_hook` and it defers nested objects that can't be decoded by the current `_decode_object` logic. 

- Other possible implementations
`from_json(_file)` and `from_config` are a bit tricky. Because the parameter or option files could be prepared in alternate ways (like manually), the files may or may not be strictly formatted and have all the required keys. Therefore, I didn't put these methods in the base class.

rwxayheee:
Usage note at fdc08f7 (same in 252bc95)

### 1. Export to JSON: Without explicitly naming the encoder
This is a changed usage, aiming to help ensure the correct encoding logic is always used. 

Before
```py
with open(filename, "w") as f:
    json.dump(polymer, f, cls=PolymerEncoder)
```
After
```py
with open(filename, "w") as f:
    f.write(polymer.to_json())
```

### 2. Create from JSON: With specific decoder from subclass 
This is relatively unchanged from the original implementation (which also requires the name of expected class). The name of the function is now shorter. 

Before
```py
decoded_polymer = json.loads(
        json_str, object_hook=polymer.polymer_json_decoder
    )
```
After
```py
decoded_polymer = Polymer.from_json(json_str)
```

### 3. Override the "Correct" decoder/encoder
It *seems* possible to override the corresponding decoder/encoder: 
```py
starting_molsetup = populated_rdkit_molsetup
# full JSON dump (correct) 
json_str = starting_molsetup.to_json() 
# partial JSON dump, as the RDKitMoleculeSetup is an extended MoleculeSetup
json_str2 = json.dumps(starting_molsetup, default=MoleculeSetup.json_encoder, skipkeys=True) 

# creates RDKitMoleculeSetup (correct)
decoded_molsetup = RDKitMoleculeSetup.from_json(json_str)
# creates a MoleculeSetup
decoded_molsetup2 = json.loads(json_str, object_hook = MoleculeSetup.json_decoder)
# creates a MoleculeSetup
decoded_molsetup3 = MoleculeSetup.from_dict(json.loads(json_str))
```

### 4. Lenient Parsing (from_dict, from_json_file, from_config)
`from_dict` is exactly `json_decoder` but allows excess keys. 

`from_json_file` and `from_config` are currently only implemented for certain classes, that may come from a manually written option or parameter file. 

### 5. Changes of certain keys
To re-use the decoder/encoder, there's some minor change in the spelling of expected keys. They were changed to be consistent with the existing JSON files in the meeko/data directory. 

```py
class ResidueTemplate:
  expected_json_keys = {"mol", "link_labels", "atom_name"} # was atom_names
```

```py
class ResiduePadder:
    expected_json_keys = {
        "rxn_smarts",
        "adjacent_res_smarts", # was adjacent_smarts
        "auto_blunt",
    }
```

@diogomart please let me know if you like this. It does not introduce new features but this has been a good experience for me to review the JSON-bound properties. I understand more about the data structure now.

meeko/utils/jsonutils.py
```diff
@@ -74,3 +93,74 @@ def string_to_tuple(input_string: str, element_type: type = str):
...
+        raise NotImplementedError("Subclasses must implement _decode_object.")
+
+    # Inheritable JSON Interchange Functions
+    @classmethod
+    def json_decoder(cls, obj: dict[str, Any], check_keys: bool = True):
+        # Avoid using json_decoder as object_hook for nested objects
+        if not isinstance(obj, dict):
+            return obj
+        if check_keys and not cls.expected_json_keys.issubset(obj.keys()):
+            return obj
```
diogomart, line 118: I know that the original code had this already, but we should try to raise an error instead of returning the input `obj`. Returning the input `obj` makes errors very hard to find as it just fails silently. Unless there is a purpose in returning the input `obj`.

meeko/polymer.py
```diff
@@ -2478,14 +2599,53 @@ class ResidueTemplate:
...
+        if deserialized_mol: 
+            mol_smiles = rdkit.Chem.MolToSmiles(deserialized_mol, canonical=False)
+        # if dry json (data) is supplied
+        else:
+            mol_smiles = obj.get("smiles")
+
+        link_labels = convert_to_int_keyed_dict(obj.get("link_labels"))
+
+        # Construct a ResidueTemplate object
+        residue_template = cls(mol_smiles, None, obj.get("atom_name"))
```
diogomart: Renaming `atom_names` to `atom_name` makes it fail to read polymer JSON written by meeko v0.6.1. I thing this was to reuse the `from_dict` function from the base class to load the default templates, which is great, but we must be able to read polymer JSON from v0.6.1. I'll push a (failing) test for this.

I'm also concerned about the `obj.get(key)` vs `obj[key]`. I think it's better to have a missing key error, which is easy to debug, than some trouble with an **un**expected None with a more complex error traceback.

diogomart (changes requested):

meeko/utils/jsonutils.py
rwxayheee: This was done intentionally. The mismatched must be deferred (delayed) here as object_hook, to avoid decoding nested objects. The check of keys is usually put in class-specific `_decode_object`.

rwxayheee: My bad, I said two things wrong in the comment above (I removed it already). Correction: (1) The check of returned object must be placed else where, if `json_decoder` needs to be used as `object_hook`. Otherwise, nested objects will be decoded with the same object_hook. (2) the check is deferred to the higher level function `from_json` in base class, not class-specific `_decode_object`

rwxayheee: The code structure is explained here: 
https://github.com/forlilab/Meeko/pull/292#issuecomment-2555925567
> Function `_decode_object` focuses on the decoding logic, while `from_json` is on a higher level and is more for error handling. On the *intermediate level*, there's `json_decoder` as object_hook and it defers nested objects that can't be decoded by the current _decode_object logic.

meeko/polymer.py
rwxayheee: `obj.get()` here are being forgiving for (manually prepared) templates that are missing properties. Could we allow missing keys ("link_labels" and/or "atom_name") in template files?

rwxayheee: The keys are checked in `from_json`. The missing or excess keys will be logged from there. If it causes a failure of object creation it will throw an RuntimeError, after attempting to create the obj. Otherwise it will just be logged. I use `obj.get()` for some keys (? maybe not all, that's bad - I should have made these more visible and consistent) in the decoding logic to allow lenient parsing. In other words, the error handling is in `from_json`. The class-specific decoding function `_decode_object` is generally forgiving

rwxayheee: Do you prefer an early return (no attempt to create obj at all) for missing keys? Currently the mismatched keys are logged if non-critical

rwxayheee:
Hi @diogomart 
Thanks for the review. I replied in comments. But here are two quick things I wanted to check with you: 

When we're creating obj from JSON: Do you prefer an early return (no attempt to create obj at all) for missing keys? Currently the mismatched keys are logged if non-critical. 

The use of `obj.get()` might be a bit inconsistent. I will double-check. But with the current code structure I'm personally ok with always using `obj.get()`, because I'm forgiving mismatched keys unless they cause a problem. Is this still ok or not ok? 
I want to change all to `obj.get()`. but maybe I'm lacking the knowledge of what could go wrong in this situation..

rwxayheee:
At present, I use `obj.get()` in very few cases (8 occurrences). There are two main situations: 

(1) The keys are optional for lenient parsing. 

(2) The keys correspond to a computed attribute, not an initializer attribute. Thus, the keyword is optional for the `__init__` function. But when I started marking the computed attribute vs. initializer attributes, I wanted to allow creation of objects from JSON in different stages of their lifecycle (processed or unprocessed). Therefore, I want to allow more `obj.get()`

meeko/utils/jsonutils.py
```diff
@@ -74,3 +93,74 @@ def string_to_tuple(input_string: str, element_type: type = str):
...
         return tuple(input_string)
+
+class BaseJSONParsable:
+
+    # Define in Individual Subclasses 
+    expected_json_keys: Optional[frozenset[str]] = None
+
+    @classmethod
+    def json_encoder(cls, obj):
+        raise NotImplementedError("Subclasses must implement json_encoder.")
```
rwxayheee: `@abstractmethod` (happens to be a github username..) is better here. If we are doing a revision I will change it

rwxayheee:
- [x] Support jsons with old syntax (attribute names)
- [x] Review use of `obj.get()`
- [x] Implements abstract method instead of not implemented error

meeko/utils/jsonutils.py
diogomart: Makes sense - thanks for clarifying.

meeko/polymer.py
diogomart: As discussed, as as you added in comment below, just need to keep backwards compatibility.

meeko/utils/jsonutils.py
diogomart: sounds good!

rwxayheee:
Changes since [681a795](https://github.com/forlilab/Meeko/pull/292/commits/681a7958f60bff7642803178d18c62833cb539f7)

* Enabled support for jsons with old keys (attribute names)
old key: `atom_names` new key: `atom_name`
This change affected the decoding logics of class `Monomer`, and `ResidueTemplate`. 

old key: `adjacent_smarts` new key: `adjacent_res_smarts`
This change affected the class `ResiduePadder`. 

To address this, I introduced a new class function, `access_with_deprecated_key`. It is inheritable from the base class. Additionally, it must be used together with lenient parsing, which is `<class name>.json_decoder(check_keys=False)`. 

* Eliminated use of `obj.get()`
This is now only used when necessary (for parsing of alternate formats). 

* Implemented abstract method instead of not implemented error


Sorry there were two changes I forgot to add in [8273d5f](https://github.com/forlilab/Meeko/pull/292/commits/8273d5f62ce5cb52f57a6880236530e8017eaec5). 

@diogomart This is ready, please let me know if it works! Thanks for your time and kind advice in advance

diogomart:
I think the try/except statements in `from_json()` obfuscate error messages. Consider the following example
```python
from meeko.molsetup import Atom

a = Atom(42)
b = Atom.from_json(a.to_json())
```
If I remove all try/except and just execute the code, I get to know that the error occured here:
```
  File "/home/diogom/Work/code/Meeko/meeko/molsetup.py", line 249, in _decode_object
    pdbinfo = PDBAtomInfo(*obj["pdbinfo"])
```
but the try/except wrap the error traceback and it's harder to find the problem.

Is there an advantage in the try/except?

rwxayheee:
Hi @diogomart 

Can you show which try/except wraps the traceback? It's possible to get the traceback while still using try/except. There's a nested try/except to distinguish different types of errors (see comments in code). But i'm not sure if this is the best approach to handle this. 

Can you suggest an alternate approach?

rwxayheee:
We discussed a similar issue before, personally I don't have the habit of printing traceback (so i forgot to take care of this, I should though..), but I recall that we agreed on doing something like this, for the PolymerCreationError: 

https://github.com/forlilab/Meeko/blob/07a38f8bac4aa9eef4ef30aa8e67475b15d517aa/meeko/polymer.py#L484-L495

Would you like the same handling, or do you prefer just not using try/except? To trace error raised from the decoding/encoding of a nested object it might need a little more work, but I could try

diogomart:
I would just remove the try/except in from_json() in jsonutils.py, then it looks like this (about half the size)
```python
    @classmethod
    def from_json(cls, json_string: str):
        # Log mismatched keys if non-critical 
        obj_dict = json.loads(json_string)
        actual_keys = obj_dict.keys()
        if actual_keys != cls.expected_json_keys:
            missing_keys = cls.expected_json_keys - actual_keys
            extra_keys = actual_keys - cls.expected_json_keys
            logging.warning(
                f"Key mismatch for {cls.__name__}. "
                f"Missing keys: {missing_keys}, Extra keys: {extra_keys}."
            )
        obj = json.loads(json_string, object_hook=cls.json_decoder)
```

diogomart:
Yeah we could re-raise the error, but I think that's just more boilerplate code for the same kind of traceback.

For the polymer construction we are catching more errors to provide more elaborate messages to the user. And to report all failures (to match templates) at once, then the user can act on all of them instead of having to run mk_prepare_receptor.py over and over again. But that is not a feature we need here.

rwxayheee:
We can just remove try/except, but it will not be able to distinguish different types of error 

```
But that is not a feature we need here.
```

(I might have written a request in an open issue for this, but couldn't find it at the moment.. maybe I'm the only one who wants this, so I implemented it) 

Try/except is always over-handling but I think it helps to know the general type of problem before exposing the traceback. I'm ok with no try/except as it requires minimal changes. But I can also do more work to print the traceback, if that's helpful. 
We can talk more when we meet!

rwxayheee:
I think all json-bound errors should be raised at from_json (because errors from lower level or nested object are deferred here). But let me think about it a bit more. The current state of this PR is also not the best way to handle this

diogomart:
Yeah, if you want the messages from the try/except keep the current try/except code, but add the detailed traceback by re-raising the error (or something similar) so we can more easily debug things in the future. In the example I posted above, the line we really need to know about is 
```
  File "/home/diogom/Work/code/Meeko/meeko/molsetup.py", line 249, in _decode_object
    pdbinfo = PDBAtomInfo(*obj["pdbinfo"])
```
we must still expose this detailed traceback.

diogomart:
The simplest way might be to re-reaise, by adding, for example, `from decoded_error`
```python
            except ValueError as decoder_error:
                raise RuntimeError(
                    f"An error occurred when creating {cls.__name__} from JSON."
                    f"Error: {decoder_error}"
                ) from decoder_error
```

diogomart:
@rwxayheee Nevermind... the traceback is already printed in the PR as is, I missed while testing and got confused. Please ignore this last few messages, sorry.

diogomart:
When trying to decode with a missing attribute we get an error as expected:
```python

from meeko.molsetup import Bond
import json

bond = Bond(42, 43, rotatable=True)
json_bond = bond.to_json()
dict_bond = json.loads(json_bond)
dict_bond.pop("rotatable")
try:
    decoded_bond = Bond.from_json(json.dumps(dict_bond))
    print("Successfully decoded Bond without rotatable attribute")
except Exception as err:
    print("Failed at decoding Bond without rotatable attribute")
    print(err)
```
We popped the "rotatable" attribute out of the JSON, so we get an error. All good here.

However, if we do the same within a molecule setup, then we get no error:

```python
from meeko import RDKitMoleculeSetup
from scrubber import Scrub
from rdkit import Chem
import json


scrub = Scrub()

butanol = Chem.MolFromSmiles("OCCC")
butanol = scrub(butanol)[0]
molsetup = mk_prep(butanol)[0]
for bond_id, bond in molsetup.bond_info.items():
    print(f"{bond_id=} {type(bond).__name__}")

data = json.loads(molsetup.to_json())
for bond_id, bond in data["bond_info"].items(): 
    bond.pop("rotatable")
decoded_molsetup = RDKitMoleculeSetup.from_json(json.dumps(data))

print("Bonds in decoded molsetup are dict, not type Bond")
for bond_id, bond in decoded_molsetup.bond_info.items():
    print(f"{bond_id=} {type(bond).__name__}")
```

This prints
```
bond_id=(0, 1) Bond
bond_id=(1, 2) Bond
bond_id=(2, 3) Bond
bond_id=(0, 4) Bond
bond_id=(1, 5) Bond
bond_id=(1, 6) Bond
bond_id=(2, 7) Bond
bond_id=(2, 8) Bond
bond_id=(3, 9) Bond
bond_id=(3, 10) Bond
bond_id=(3, 11) Bond
Bonds in decoded molsetup are dict, not type Bond
bond_id=(0, 1) dict
bond_id=(1, 2) dict
bond_id=(2, 3) dict
bond_id=(0, 4) dict
bond_id=(1, 5) dict
bond_id=(1, 6) dict
bond_id=(2, 7) dict
bond_id=(2, 8) dict
bond_id=(3, 9) dict
bond_id=(3, 10) dict
bond_id=(3, 11) dict
```
We should still be getting an error.
```

**Comment by rwxayheee** [comment]
Hi @diogomart 
Thanks for the review. Sounds like we're able to see the error raised from a nested object. I recall that we should have the traceback from the point where the error was raised. There's room to improve if other people need more detailed error, for now I will not modify the current PR at the moment, but please let me know if there's anything else I can do. 

One last note on this is that this PR doesn't include #254. After we merge this, I can resubmit another PR in place of #254.

**Comment by diogomart** [comment]
I already merged this PR into 254, no worries about that one. But could you take a look at why removing an attribute from a JSON string and passing it to the from_json constructor of the respective class raises an error, for example removing the `rotatable` attribute of a bond and passing the JSON string to `Bond.from_json()`, but we don't get an error if we remove the `rotatable` attribute from the bonds of a molecule setup's JSON and pass it to `RDKitMoleculeSetup.from_json()`.
edit: see https://github.com/forlilab/Meeko/pull/292#issuecomment-2581152557 above

**Comment by rwxayheee** [comment]
@diogomart I get what you mean now, sorry I misunderstood the comment. 
There's some quick way to fix (replacing json_decoder by from_dict or from_json with options), but allow me to reconsider the check_keys option. I'm not very sure if I implemented it in the way I want..

**Comment by rwxayheee** [comment]
Hi @diogomart 
[761e829](https://github.com/forlilab/Meeko/pull/292/commits/761e8299a07b4bddb341a5b544390c03a97be7a9) replaced all `json_decoder` by `from_dict` to decode a nested object. This way any problems with decoding the nested object will be directly raised from where they are from. In the example you provided, the missing key `rotatable` triggers an error from here: 

```
  File "/Users/amyhe/Desktop/0_forks/Meeko/meeko/molsetup.py", line 306, in _decode_object
    rotatable = obj["rotatable"]
KeyError: 'rotatable'

During handling of the above exception, another exception occurred:

  File "/Users/amyhe/Downloads/test_json_pr.py", line 21, in <module>
    decoded_molsetup = RDKitMoleculeSetup.from_json(json.dumps(data))
  File "/Users/amyhe/Desktop/0_forks/Meeko/meeko/utils/jsonutils.py", line 168, in from_json
    raise RuntimeError(
RuntimeError: Unable to load JSON string for RDKitMoleculeSetup. Error: 'rotatable'
```

It relies on directly accessing a keyword from dict, not `obj.get`. As you suggested before, `obj.get` should be avoided as much as possible. 

This is the minimal fix (without implementing more sophisticated error passing functions) I can think of. I will stay with the current usage of `check_keys` as it prevents unexpected decoding of nested objects.

rwxayheee:
Just one more note - I feel it would be helpful to have some function for each class to validate if the nested objects are the correct type. The type-validation function is not strictly JSON-bound, could be useful in a more general way. 
Currently it's possible in Python to assign anything (not the expected type) to a nested object, not just from JSON. 

Happy to work on it in a new issue or PR!

diogomart:
The real problem might be.... Python itself with dynamic typing.

diogomart (approved):

rwxayheee:
Thanks for the approval! Merging to develop

PR #301, opened 2024-12-27
Ranking passed templates by H_excess if tied with H_missing

### LINKED ISSUE #300: Triply protonated N-terminal amines should not be modeled blunt
Input hydrogens are being ignored, which is not desired behavior. It seems that all terminal amino-acids are modeled blunt. The desired default behavior is the following, depending on the number of hydrogens on the amine (subtract one for proline):
 - 0 H -> blunt
 - 1 H -> blunt
 - 2 H -> ideally -NH2 but I don't think we have templates for that, this is very rare, either -NH2 or blunt is OK
 - 3 H -> -NH3+

rwxayheee:
This is a quick fix for #300. Passed current tests (in develop) and works for the triply protonated N-terminal residues. It implements the additional refinement of passed templates (previously forgiven with excess H) for minimal excess H. 

My previous PR #211 caused the unexpected behavior. This is an amend to handle explicitly protonated linking atoms in the corect way. 

@diogomart Would this be ok as a temporary fix? Sorry for implementing the bug ><


rwxayheee:
Please hold off for a short while. I would like to include two additional bug fixes before finalizing this PR. They were included in #277 but I gave up on that PR. Only need very minimal changes.

rwxayheee:
Added two commits that were part of #277. More detail is explained in #277. 

- Bug fix for #280. This is a quick fix without changing other parts of the code. #277 has more efficient handling of covalent docking results, but requires too many changes. 

- Improvement for #231. This does not change the behavior of template matching. 

Added example file from #300 as a test.

diogomart (approved):
This looks great. Thank you. Sorry for the long delay.

PR #306, opened 2025-01-09
added function stitch from PR188, with options `residues_to_add` and `bonds_to_use`

rwxayheee:
This is a modified version of PR188. 

It added `bonds` attribute to all Polymer instances. And introduced a function `stitch` to output rdkit mol(s) from one or more selected monomers from a polymer. 

The major modification is to enable custom `residues_to_add` and `bonds_to_use`, which might be useful when a monomer to stitch is coming from a rdkit mol (covalent ligand). See the following example: 

```py
# example usage
from meeko import ResidueChemTemplates
from meeko import Polymer, Monomer
from meeko.polymer import find_inter_mols_bonds

import prody
from rdkit import Chem

# construct any input polymer
pdb_fn = "5tmn.pdb"
input_obj = prody.parsePDB(pdb_fn, altloc="all")
templates = ResidueChemTemplates.create_from_defaults()
polymer = Polymer.from_prody(
                input_obj,
                templates,
                mk_prep
                )

def monomer_from_mol(mol): 
    num_atoms = mol.GetNumAtoms()
    mapidx_to_raw = {i: i for i in range(num_atoms)}
    return Monomer(mol, mol, mapidx_to_raw, # initialize with the same raw_input_mol and rdkit_mol
                   atom_names=["X"]*num_atoms, input_resname="XXX") # required proprties by polymer.to_pdb()

# construct a new monomer from rdkit mol
polymer.monomers["E:0"] = monomer_from_mol(polymer.monomers["E:1"].rdkit_mol)

# find and index bonds
bonds = find_inter_mols_bonds({
    "E:0": (polymer.monomers["E:0"].rdkit_mol, polymer.monomers["E:0"].input_resname), 
    "E:2": (polymer.monomers["E:2"].rdkit_mol, polymer.monomers["E:2"].input_resname), 
})

# stitch specified residues by specified bonds
polymer_mol = polymer.stitch(residues_to_add=["E:2", "E:0"], bonds_to_use=bonds)

# output for inspection and comparison
output_file = "output.pdb"
Chem.MolToPDBFile(polymer_mol, output_file)

fn = "topdb.pdb"
with open(fn, "w") as f:
    f.write(polymer.to_pdb())
```

### DIFF
--- a/meeko/polymer.py
+++ b/meeko/polymer.py
@@ -907,6 +907,7 @@ class Polymer:
             invmap2 = {j: i for i, j in monomer2.mapidx_to_raw.items()}
             _bonds[key] = [(invmap1[b[0]], invmap2[b[1]]) for b in bond_list]
         bonds = _bonds
+        self.bonds = bonds
 
         # padding may seem overkill but we had to run a reaction anyway for h_coord_from_dipep
         padded_mols = self._build_padded_mols(self.monomers, bonds, padders)
@@ -919,6 +920,72 @@ class Polymer:
             self.parameterize(mk_prep)
 
         return
+
+    def stitch(self, residues_to_add: Optional[set[str]] = None, 
+               bonds_to_use: Optional[dict[tuple[str], list[tuple[int]]]] = None):
+        """returns a single rdkit molecule that results from adding bonds
+            between every chorizo residue. It may contain multiple fragments
+            if there are multiple chains or gaps. 
+
+            Optionally, specify a set of residue IDs for stitching.
+            Defaults to stitching all monomers. 
+
+            Optionally, specify a dict for bonds to use, 
+            Defaults to stitching using all available bonds in polymer. 
+            key format: (res_id_1, res_id_2)
+            value format: [(atom_idx_1, atom_idx_2), ]
+            same format as output from function find_inter_mols_bonds, 
+            but the indices need to based on rdkit_mol. 
+        """
+        
+        # stitching all monomers by default
+        residues_to_add = residues_to_add or self.monomers.keys()
+        residues_to_add = set(residues_to_add)
+        bonds_to_use = bonds_to_use or {k: v for k, v in self.bonds.items() if k[0] in residues_to_add and k[1] in residues_to_add}
+
+        # verify if requested monomers are valid (have rdkit_mol)
+        valid_monomers = set(self.get_valid_monomers().keys())
+        invalid_monomers = residues_to_add - valid_monomers
+        if invalid_monomers: 
+            raise ValueError(f"Residue IDs not in valid monomers: {invalid_monomers}")
+        
+        # initialize mol and residue/bond tracking
+        mol = Chem.Mol()
+        residues_added = {}
+        bonds_spent = set()
+        
+        # add residues and get offset in order
+        offset = 0
+        for r_id in residues_to_add:
+            res = self.monomers[r_id]
+            residues_added[r_id] = offset
+            offset += res.rdkit_mol.GetNumAtoms()
+            mol = Chem.CombineMols(mol, res.rdkit_mol)
+
+        # add bonds
+        edit_mol = Chem.EditableMol(mol)
+        for bond_key, bond_list in bonds_to_use.items():
+            if bond_key in bonds_spent:
+                continue
+            r1, r2 = bond_key
+            if r1 in residues_added and r2 in residues_added:
+                bonds_spent.add(bond_key)
+                for bond in bond_list: 
+                    i, j = bond
+                    edit_mol.AddBond(
+                        i + residues_added[r1],
+                        j + residues_added[r2],
+                        order=Chem.rdchem.BondType.SINGLE
+                    )
+        mol = edit_mol.GetMol()
+        
+        # review added bonds and residues
+        if len(bonds_spent) != len(bonds_to_use):
+            raise RuntimeError("nr of bonds added differs from bonds to use")
+        if len(residues_added) != len(residues_to_add):
+            raise RuntimeError("nr of residues added differs from residues to add")
+        
+        return mol
 
     @classmethod
     def from_pdb_string(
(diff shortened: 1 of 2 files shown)


rwxayheee:
Just commenting for the future record: This could be a useful function not just to combine covalent ligand and residue. A generalized stitching function might be a handy tool when users want to forgive elongated bonds that are not formed when processing with the parser.

diogomart:
Thank you Amy! Finally started testing, here's what I found
 - by default, when no residue IDs are passed, it should default to `self.monomers.get_valid_residues()` to avoid an erroring due to invalid residues
 - the polymer should save the `bonds` attribute when serializing to JSON
 - the polymer should optionally read the `bonds` attribute when reading from JSON, optionally for backwards compatibility. We can compute the bonds if they are missing in the JSON using the `find_inter_mols_bonds` function.
 - we should, in a different PR, add a check for JSON serialization in which we check if newly added attributes are being tested. The idea is to catch new attributes, like the `bonds` herein, that are added to the classes are in fact being tested. Otherwise, we are dependent on remembering to add them to the encoder/decoder.

rwxayheee:
Hi @diogomart 

Thanks for the comments. To be completely honest, I might not fully understand the original motive you created the stitch function, so I don't fully understand the comments.. This function currently is a pipeline-like function (so not a critical dependency). If this is not going in the direction you wanted, you could take over and close this PR. 

Currently in this PR, this function (polymer.stitch) returns an RDKit molecule. I considered this function for the following use cases: (1) making covalently modified residues and templates, (2) constructing molecules with manually specified (potentially distorted) bonds. 

Regarding your comments: 


> by default, when no residue IDs are passed, it should default to self.monomers.get_valid_residues() to avoid an erroring due to invalid residues

I see. 


> the polymer should save the bonds attribute when serializing to JSON
> the polymer should optionally read the bonds attribute when reading from JSON, optionally for backwards compatibility. We can compute the bonds if they are missing in the JSON using the find_inter_mols_bonds function.

This doesn't seem to be related to this PR (this function is not a polymer constructor) so I will try to respond on the project level: 

You seem to be commenting on the JSON-related functions about the polymer class, and it seems like you want some custom behaviors on certain classes with certain attributes. 

The current JSON-related base functions are like the basis (similar to a file parser) and they don't convey any chemically meaningful conventions. To achieve the goals, you could make a pipeline-like function to construct a polymer from json, very much like other polymer constructors we have (from PDB, from ProDy), that follows certain desired expected behaviors. 


> we should, in a different PR, add a check for JSON serialization in which we check if newly added attributes are being tested. The idea is to catch new attributes, like the bonds herein, that are added to the classes are in fact being tested. Otherwise, we are dependent on remembering to add them to the encoder/decoder.

Here are some options to consider: 

(a) Parse classes using [ast](https://docs.python.org/3/library/ast.html), and calculate the differences in attributes against the old attribute assignment -> make it a pre-commit check and flags newly added or modified attributes

(b) Establish a project schema -> use a schema-first code generator to write tests against new edits, specifically against modified attributes

(c) Track test coverage (coverage.py, Codecov)

diogomart:
There's a new attribute: `Polymer.bonds`. This PR is adding a new attribute, so it should make sure that this new attribute is in line with what is expected of other attributes. Thus, `Polymer.bonds` should be JSON serialized and deserialized like any other attribute. This way if we create a Polymer from scratch, or if we dump in to JSON and read it again, we can still call the stitch function. We also want to ensure backwards compatibility and be able to stitch Polymer that were serialized with v0.6.1.

rwxayheee:
> There's a new attribute: Polymer.bonds.

ahhhhh, I see. It was added in #188. I found it available and was using it to set the default bonds_to_use, then I forgot this was a new attribute.... 
Do you want to store the bonds within polymer? I see it could be useful. But it's not strictly necessary for this function. The bonds can be re-calculated anytime as long as we are keeping raw_input_mol for each residue.

diogomart:
I think it's fine either way. We are keeping the raw input mols. Re-calculating is always necessary for backwards compatibility. Re-calculating might also be easier to maintain once we start editing the polymer more aggressively, by deleting, building, or mutating residues, because it's safer to calculate the bonds based on the current state of the polymer than to assume the existing bond attribute is being updated with structural edits. So I guess it may be better to just compute the bonds on the fly.

rwxayheee:
> I think it's fine either way. We are keeping the raw input mols. Re-calculating is always necessary for backwards compatibility. Re-calculating might also be easier to maintain once we start editing the polymer more aggressively, by deleting, building, or mutating residues, because it's safer to calculate the bonds based on the current state of the polymer than to assume the existing bond attribute is being updated with structural edits. So I guess it may be better to just compute the bonds on the fly.

Thanks! I made the changes so this function doesn't rely on the bonds attribute. But if we think it's useful, we can add the attribute back in future. And we can look at ways to properly track the modifications in attributes. 

In this specific function, the polymer's bonds attribute cannot always be used as the default, because we need the number of bonds supplied to be exactly the number of bonds being used to stitch the monomers. This means the Polymer's bonds attribute may be the reasonable default, but only when we are stitching all valid monomers in the polymer. 

Currently, the default bonds are re-calculated based on `residues_to_add`.

diogomart (changes requested):
Could you add this test to `polymer_creation_test.py`

```python
def test_stich_polymer():
    with open(disulfide_adjacent, "r") as f:
        pdb_text = f.read()
    polymer = Polymer.from_pdb_string(
        pdb_text,
        chem_templates,
        mk_prep,
    )
    adjacent_disulfide = Chem.MolFromSmarts("S1CCNCCCS1")
    disulfide_then_proline = Chem.MolFromSmarts("CSSCCC(=O)N1CCCC1")
    stitched_mol = polymer.stitch()
    assert stitched_mol.HasSubstructMatch(adjacent_disulfide)
    assert stitched_mol.HasSubstructMatch(disulfide_then_proline)
    # after serialization
    polymer = Polymer.from_json(polymer.to_json())
    stiched_mol = polymer.stitch()
    assert stitched_mol.HasSubstructMatch(adjacent_disulfide)
    assert stitched_mol.HasSubstructMatch(disulfide_then_proline)
```

diogomart (approved):
Thanks!

PR #315, opened 2025-01-30
Remove all uppercase types CL and BR from GPF file

diogomart:
In MacOS filesystem, which is case-insensitive, they are the same as the mixed case Cl/Br counterparts, causing confusion.


atillack (approved):
@diogomart Yes, please. This is a great idea going forward.

rwxayheee (approved):

PR #318, opened 2025-02-19
Update required Python version

pslacerda:
Dropped the Python version upper limit as is usual to actively maintained projects. It is now installable in Python 3.13.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ setup(
     include_package_data=True,
     zip_safe=False,
     install_requires=['numpy>=1.18', "rdkit", "scipy", "gemmi"],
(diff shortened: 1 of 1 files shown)


rwxayheee:
Hi @pslacerda 
We restrict the version for a reason, see: 
https://meeko.readthedocs.io/en/release-doc/installation.html#support-for-python-3-12

Part of this project depends on ProDy. One important feature that relies on ProDy is the preparation for covalent (tethered docking). For a ProDy-independent version, see this: 
https://github.com/forlilab/Meeko/pull/277
However this is never merged, and not up to date to many updates in develop.

PR #325, opened 2025-03-14
Small Bugfixes in closed PR #259

rwxayheee:
1- Allow single atom residues in template generation

~~2- Bugfix for option `--write_pdb` in CLI script mk_prepare_receptor~~

This is not a bug, but a behavior I didn't fully understand. 

Before: 

```
(mk_dev) amyhe@Amys-MBP ~/Downloads % mk_prepare_receptor.py -i three_res.pdb -o three_res_out --write_pdb
@> 4412 atoms and 1 coordinate set(s) were parsed in 0.02s.
  File "/Users/amyhe/micromamba/envs/mk_dev/bin/mk_prepare_receptor.py", line 33, in <module>
    sys.exit(load_entry_point('meeko', 'console_scripts', 'mk_prepare_receptor.py')())
  File "/Users/amyhe/Desktop/0_forks/Meeko/meeko/cli/mk_prepare_receptor.py", line 630, in main
    fn = args.write_pdb[0]
IndexError: list index out of range
```

3- Corrected polymer.to_pdb() that wrote element in columns 78-79

Before we wrote: 

```
ATOM      1  CL  CL  B1464      36.135  12.381  -3.884                       Cl 
```

After (expected):

```
ATOM      1  CL  CL  B1464      36.135  12.381  -3.884                      Cl 
```

### DIFF
--- a/meeko/cli/mk_prepare_receptor.py
+++ b/meeko/cli/mk_prepare_receptor.py
@@ -634,7 +634,7 @@ def main():
         with open(fn, "w") as f:
             f.write(polymer.to_pdb())
         written_files_log["filename"].append(fn)
-        written_files_log["description"].append("receptor")
+        written_files_log["description"].append("processed receptor PDB")
     
     if args.write_pdbqt is not None:
         if args.write_pdbqt:
--- a/meeko/cli/mk_prepare_receptor.py
+++ b/meeko/cli/mk_prepare_receptor.py
@@ -630,7 +630,7 @@ def main():
         if args.write_pdb:
             fn = args.write_pdb[0]
         else:  
-            fn = str(outpath) + ".pdb"
+            raise ValueError("--write_pdb requires a filename")
         with open(fn, "w") as f:
             f.write(polymer.to_pdb())
         written_files_log["filename"].append(fn)
--- a/meeko/cli/mk_prepare_receptor.py
+++ b/meeko/cli/mk_prepare_receptor.py
@@ -627,7 +627,10 @@ def main():
         written_files_log["description"].append("parameterized receptor")
     
     if args.write_pdb is not None:
-        fn = args.write_pdb[0]
+        if args.write_pdb:
+            fn = args.write_pdb[0]
+        else:  
+            fn = str(outpath) + ".pdb"
         with open(fn, "w") as f:
             f.write(polymer.to_pdb())
         written_files_log["filename"].append(fn)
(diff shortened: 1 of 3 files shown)


meeko/cli/mk_prepare_receptor.py
```diff
@@ -627,7 +627,10 @@ def main():
         written_files_log["description"].append("parameterized receptor")
     
     if args.write_pdb is not None:
-        fn = args.write_pdb[0]
+        if args.write_pdb:
+            fn = args.write_pdb[0]
+        else:  
+            fn = str(outpath) + ".pdb"
```
diogomart: This was done on purpose to prevent accidentally overwriting the input file. Usually I spend some effort preparing a model, and then save this in a PDB file named, "target.pdb" or similar. Then I write "target.pdbqt", "target.json" with `-o target -p -j`. If I pass also `--write_pdb` it will overwrite the input "target.pdb". Forcing the user to type this explicitly `--write_pdb target.pdb` is safer, I think.

meeko/polymer.py
```diff
@@ -1831,7 +1831,7 @@ def to_pdb(self, new_positions: Optional[dict]=None):
 
         pdbout = ""
         atom_count = 0
-        pdb_line = "{:6s}{:5d} {:^4s} {:3s} {:1s}{:4d}{:1s}   {:8.3f}{:8.3f}{:8.3f}                       {:2s} "
+        pdb_line = "{:6s}{:5d} {:^4s} {:3s} {:1s}{:4d}{:1s}   {:8.3f}{:8.3f}{:8.3f}                      {:2s} "
```
diogomart: This is OK but we could use "{:>2s}" to write-align the single-char elements

diogomart (changes requested):

meeko/polymer.py
rwxayheee: > This is OK but we could use "{:>2s}" to write-align the single-char elements

Is right-align (?) the preferred behavior for single-char elements? Do you know if other software usually does this? I did the fix because element was written in the wrong number of columns

meeko/cli/mk_prepare_receptor.py
rwxayheee: That makes sense, I was misusing the option

diogomart (approved):

meeko/polymer.py
diogomart: I see you fixed, thanks. Just for the record, it is right-justified:
https://www.cgl.ucsf.edu/chimera/docs/UsersGuide/tutorials/pdbintro.html

PR #328, opened 2025-03-18
Use warnings or logger rather than print

k-ujihara:
This PR changes print statements to logging or warnings class.


diogomart:
Thanks! This looks great. I think we might need to configure a root logger from the command line scripts (in meeko/cli). I'll look into it soon.

rwxayheee (approved):
This looks good, approve and thank you @k-ujihara for the nice work!

diogomart:
My only concern is that the downloaded CCD templates are not printed anymore. But I guess that's OK because we now cache them after #287. Thus, I'm merging. Thank you all.

PR #332, opened 2025-03-26
Making Ala Gly Pro flexible doesn't raise error, simply writes to rigid part

diogomart:


rwxayheee:
This might write an partly empty flexible receptor file like the following: 

```
BEGIN_RES ALA A 19
REMARK INDEX MAP
ROOT
ENDROOT
END_RES ALA A 19

```

Is this the intended behavior? Would it potentially cause a parsing error? 

Could these residues be logged and skipped instead?

diogomart:
>Could these residues be logged and skipped instead?

Yes, I like that better, I'll do that.

meeko/writer.py
```diff
@@ -523,11 +526,16 @@ def write_from_polymer(cls, polymer):
...
+                if len(graph[root]) > 1:
                     raise RuntimeError(
                         f"flexible residue {res_id} has {len(graph[root])}"
                         " rotatable bonds from root, but PDBQT is limited to 1"
                     )
+                elif len(graph[root]) == 0:
+                    logger.warning(
+                        f"flexible residue {res_id} has no movable atoms, omitting from flexres PDBQT"
+                    )
+                    continue
```
rwxayheee: Hi @diogomart, Thanks for the updates! Here it might need some small adjustment to allow writing Ala Gly Pro residues to the receptor file. With continue in L538, it directly proceeds to the next residue, and theses residues will not be written. A working structure would be something like this: 
```py
if len(graph[root]) > 1:
    # Raise an error
elif len(graph[root]) == 0:
    # Log a warning
else:
    # Do the task that we wanted to skip up to flex_pdbqt_dict[res_id]
```

diogomart: Great catch, thanks. This would have been bad. Will also add a test that would have caught this mistake.

diogomart: Actually, I think my original code is OK. Can you try to produce the bug you were expecting?

rwxayheee: When continue was used here, the residues that meet the condition (Ala Gly Pro) were not written to the rigid receptor file. This is because continue skips not only the writing-to-flexible part but also writing-to-rigid part ahead. It proceeds right to the next residue after logging the warning

rwxayheee: Continue changes the loop behavior not just escaping from the closest if. So for Ala Gly Pro they were skipped and never got a chance to make it to the writing-to-rigid part. Are you able to reproduce this, or maybe it’s something I did wrong in my part? 🥺

test/polymer_creation_test.py
```diff
@@ -122,6 +122,40 @@ def test_AHHY_all_static_residues():
...
 
+def test_AHHY_flex_residues():
+    f = open(ahhy_example, "r")
+    pdb_string = f.read()
+    polymer = Polymer.from_pdb_string(
+        pdb_string,
+        chem_templates,
+        mk_prep,
+    )
+    polymer.flexibilize_sidechain("A:2", mk_prep)
```
rwxayheee, line 133: Could you try `polymer.flexibilize_sidechain("A:1", mk_prep)`? The expected should be `len(rigid_part) == 3555`. If we use `continue` we will get 3081

rwxayheee:
Hi @diogomart 

I have an example with [efd683e](https://github.com/forlilab/Meeko/pull/332/commits/efd683ebd2e6c615ab0edd0b5927d29496f5a3df)
in case you want to take a quick look. It should be reproducible with an arbitrary structure. 

Input file
[101m.pdb.txt](https://github.com/user-attachments/files/19495876/101m.pdb.txt)

Command
`mk_prepare_receptor.py -i 101m.pdb -f A:19,15,21 -o 101m_output -p`

Rigid Output
[101m_output_rigid.pdbqt.txt](https://github.com/user-attachments/files/19495879/101m_output_rigid.pdbqt.txt)

The structure I suggested (if ... elif ... else) is a quick in-place fix I can think of, maybe not the best. If you have a different approach, you can adjust the flow and achieve the same outcome.

meeko/writer.py
diogomart: You are right... my bad, sorry. Fixed in 3c896839e66a1abefdfa4f42301c9e9dea64083b

diogomart:
Thank you! Should be good now. Also fixed the test so it _really_ would have caught the bug I introduced in efd683ebd2e6c615ab0edd0b5927d29496f5a3df

rwxayheee (approved):
This looks good, thanks!

PR #333, opened 2025-03-26
Rotatable terminal group option

althea-hansel:
Add option for making terminal groups (e.g. thiols, hydroxyls, amides) of select residues rotatable.

Performed manual testing to:
1. Confirm correct atoms are portioned between `_flex.pdbqt` and `_rigid.pdbqt` for each eligible residue
2. Confirm residue will be treated as fully flexible if specified with both `-f` and `-t`
3. Confirm error catching for ineligible residue specifications

In the process of testing that resulting docked poses are able to be re-stitched with flexres export


joanimato:
@diogomart This looks good to me. I added some examples and updated the docs.

joanimato (approved):

PR #338, opened 2025-04-11
Refactor `json_serialization_test.py` to use schema-based hierarchical tests

rwxayheee:
This is a rewrite of `json_serialization_test.py` to achieve the three major objectives: 

- Code efficiency
This PR follows the hierarchical spirit of the original JSON tests, in a schema-first way with minimal redundant codes. Hopefully this is easier for long-term maintainance and automated test coverage assessment. 

- Ensure exhaustive attribute checking
This PR exhaustively checks all attributes of all JSON-parsable classes, without hardcoding the list of attributes to check. As requested in https://github.com/forlilab/Meeko/pull/306#issuecomment-2733988070 by @diogomart, the JSON-interchange logics of new attributes must be checked. This PR demands all attributes to be evaluated in original and decoded object, unless explicitly ignored. Otherwise, pytest will throw errors. 

- Deep equality check and bug fixes
This PR implements a deep recursive equality check, instead of hardcoding how each attribute should be evaluated. Because of this, some bugs are exposed. 

(Bug fixes will be added in separate commits, in case we want to fix actual problems without major changes of the tests)

PR #339, opened 2025-04-12
Remove attribute MoleculeSetup.is_sidechain

diogomart:
`Monomer.is_movable` has superseded `MoleculeSetup.is_sidechain`.
This PR adds a test that creates a `Polymer`  from a JSON file written by v0.6.1, to guarantee backwards compatibility for future changes (the test doesn't assert anything, it only fails if there are errors).


test/json_serialization_test.py
```diff
@@ -45,6 +46,11 @@
 chem_templates = ResidueChemTemplates.create_from_defaults()
 mk_prep = MoleculePreparation()
 
+def read_v061_polymer():
```
rwxayheee: test functions needs to start with test_

diogomart: Thanks!

rwxayheee (approved):
This looks good, thanks!! Approve and merging

rwxayheee:
Thanks for providing the JSON in old version. It's very useful. I can add a full hierarchical test for it in #338

PR #341, opened 2025-04-21
(Fix #215) Update polymer/chorizo padded molecule with positions from adjacent residue when there are multiple bonds between a pair of residues

### LINKED ISSUE #215: Update polymer/chorizo padded molecule with positions from adjacent residue when there are multiple bonds between a pair of residues
This is not a priority. We currently do not rely on padding atoms having positions.

rwxayheee:


rwxayheee:
In the original code, the padded_mol did not necessarily contain all hydrogens.

Although the PR works, it requires all hydrogens, so it does not reproduce `REMARK INDEX MAP` (`mapidx_pad`). 

And therefore it also breaks test `test_export_sidechains_no_idxmap` that relies on old indices, although with new indices this PR should work. 

It seems like the tests (`test_export_sidechains_no_idxmap` and `test_AHHY_flex_residues`) may rely on the exact `REMARK INDEX MAP`.

rwxayheee:
Closing this, while it technically works, it changes atom indexing and hydrogen count in a way that could affect many things down the line.

rwxayheee:
I will give another try in the next few days. I was doing the additional `padded_mol = Chem.AddHs(padded_mol, addCoords=True)` which may be causing the difficulties. Still interested in working on this

rwxayheee:
Changed the logic in `_build_padded_mols` to reproduce exactly the index mapping between padded molecule and rdkit molecule as `develop`. 
Before 7ec6b66, the code was working for all but that part in `export_flexres.py` that uses `index_map`. 
The latest version (7ec6b66) should be working. 

Before the change, atoms added during the making of `padded_mol` were missing positions. After the change, the `padded_mol` has all hydrogens with conformations. This can be visually verified with the example file `test/polymer_data/disulfide_bridge_in_adjacent_residues.pdb`: 

```py
from meeko import Polymer, ResidueChemTemplates, MoleculePreparation
from rdkit import Chem

residue_chem_templates = ResidueChemTemplates.create_from_defaults()

input_file = "disulfide_bridge_in_adjacent_residues.pdb"
with open(input_file, "r") as f:
    pdb_string = f.read()

polymer = Polymer.from_pdb_string(
    pdb_string,
    residue_chem_templates, 
    mk_prep)

assert isinstance(polymer, Polymer), "Polymer object not created correctly"

for res_id in polymer.monomers:
    if res_id in ["G:187", "G:188"]:
        padded_mol = polymer.monomers[res_id].padded_mol
        assert padded_mol is not None, f"Padded molecule for residue {res_id} is None"
        
        writer = Chem.SDWriter(f"padded_mol_{res_id}.sdf")
        writer.write(padded_mol)
        writer.close()       
```

diogomart (approved):

PR #342, opened 2025-04-22
espaloma torsion potentials in kcal/mol, not kJ/mol

diogomart:
Thank you @rwxayheee for finding the bug


rwxayheee (approved):
looks good, thank you!

mallanos:
Thanks @rwxayheee good catch!

Issue #334, opened 2025-03-28
calling monomer.parameterize should call mk_prep.calc_flex if monomer.is_movable


diogomart:
If a monomer is made movable and then is re-parameterized, and the new parameterization has a different set of rotatable bonds, the flexibility tree should be updated automatically. Consider making `monomer.is_movable` a private attribute to hint users that chaging it manually is not enough (`Polymer.flexibilize_sidechain` should be called)

PR #343, opened 2025-04-22
Flexibilize then parameterize

diogomart:
Work in progress to address #334: so far only added a test that fails, as expected, as parameterizing a flexible monomer makes backbone bonds rotatable, which is incorrect. Goal of this PR will be to move most of the `Polymer.flexibilize_sidechain()` to `Monomer` so that calling `Monomer.parameterize` doesn't make backbone bonds rotatable.


diogomart:
The code to make a side chain flexible automatically detects backbone. An equivalent or similar algorithm was described by @atillack. Basically, only bonds that do not affect adjacent residues are rotated. To achieve that, all bonds in a graph path between atoms that have connections to adjacent residues are made non-rotatable. Counterintuitively, this can be done in the `Monomer` class as opposed to being done in the `Polymer` class because the templates define link atoms (atoms with connections to adjacent residues). An advantage is that this works for blunt residues, i.e. the last residue before a missing one, meaning that only the side chain part is made flexible.

The tricky thing is that we can call `Monomer.parameterize(new_mk_prep)` on a monomer that has been made flexible, and `new_mk_prep` can include rules for defining rotatable or non-rotatable bonds that differ from those that originally parameterized the monomer. Thus, when calling `parameterize()` on a monomer that `is_movable` (aka flexible sidechain), we must call the `flexibilize` code to update the flexibility tree and `Monomer.is_flexres_atom`.

diogomart:
Don't mind the updated constant in `espalomatyper.py`, that's already merged #342, I started off the wrong branch

diogomart:
We should test that the set of rotatable bonds (`Monomer.molsetup.bond_info[bond_id].rotatable`) and atoms in `Monomer.is_flexres_atom` is the same independently of the order of operations:
 - flexibilize -> parameterize
 - parameterize -> flexibilize
 - flexibilize -> JSON I/O -> parameterize
 - parameterize -> JSON I/O -> flexibilize

we might also want a method to deflexibilize. Simply setting the `Monomer.is_movable` flag to false doesn't reset the `Monomer.molsetup.bond_info[bond_id].rotatable` and `Monomer.is_flexres_atom` attributes to their original state. There's currently no known issues caused by this, but it's dirty.

joanimato (approved):
Make sure the conflict is resolved before merging.

Issue #331, opened 2025-03-26
monomer.molsetup_mapidx in polymer.py sometimes has class NoneType (should be dict) so fails


hrp1000:
Hi

Running this test code (pdb file is c3hrwA_.7.txt - github won't let me upload a file with ".pdb" extension) 

[c3hrwA_.7.txt](https://github.com/user-attachments/files/19472936/c3hrwA_.7.txt)

from meeko import (
    MoleculePreparation,
    Polymer,
    ResidueChemTemplates,
)
from pathlib import Path

pdb_file = Path("c3hrwA_.7.txt")
with open(pdb_file, "r") as filein:
    pdb_string = filein.read()

templates = ResidueChemTemplates.create_from_defaults()
mk_prep = MoleculePreparation()
chorizo = Polymer.from_pdb_string(pdb_string, templates, mk_prep, allow_bad_res=True)
chorizo.flexibilize_sidechain("A:15", mk_prep)

monomer.molsetup_mapidx should be a dict, _sometimes_ (for bad residues that I've allowed) it's a None (but sometimes it is a dict).

I can circumvent this if I put in a return from flexibilize_sidechain if it's a None, but I'm not sure this is a good idea... the problem also shows for A:11.

rwxayheee:
Hi @hrp1000 

The messages when you constructed the polymer: 

```
No template matched for residue_key='A:11'
tried 3 templates for residue_key='A:11'excess_H_ok=False
VAL        heavy_miss=0 heavy_excess=0 H_excess=[] bond_miss=set() bond_excess={1}
NVAL       heavy_miss=0 heavy_excess=0 H_excess=[] bond_miss=set() bond_excess={1, 4}
CVAL       heavy_miss=1 heavy_excess=0 H_excess=[] bond_miss=set() bond_excess={0, 1}

No template matched for residue_key='A:15'
tried 3 templates for residue_key='A:15'excess_H_ok=False
TRP        heavy_miss=0 heavy_excess=0 H_excess=[] bond_miss=set() bond_excess={10}
NTRP       heavy_miss=0 heavy_excess=0 H_excess=[] bond_miss=set() bond_excess={12, 4}
CTRP       heavy_miss=1 heavy_excess=0 H_excess=[] bond_miss=set() bond_excess={1, 11}

- Template matching failed for: ['A:11', 'A:15'] Ignored due to allow_bad_res.
```

[...]

PR #345, opened 2025-04-23
Raise error when trying to flexibilize a monomer that isn't valid

diogomart:
A more clear error following up on #331


diogomart:
Already in #343

PR #347, opened 2025-04-23
Forgive extra interresidue bonds

diogomart:
@pietscio is preparing receptors from @nbruciaferri  that have clashes between sidechains, and @sforli suggested to forgive excess bonds. I think this should even be made default as I can't think of a scenario in the desired behavior for a system with clashes isn't to get a warning but still process the structure. All existing tests pass. @rwxayheee can you think of a possible problem with this? We discussed some concepts in #211 but I don't remember ever discussing forgiving excess inter-residue bonds.

There's minor fixes for mk_prepare_receptor.py included herein.

### DIFF
--- a/meeko/cli/mk_prepare_receptor.py
+++ b/meeko/cli/mk_prepare_receptor.py
@@ -10,8 +10,6 @@ import sys
 import numpy as np
 
 from meeko.reactive import atom_name_to_molsetup_index, assign_reactive_types_by_index
-from meeko.polymer import residue_chem_templates_json_decoder
-from meeko import ResidueChemTemplatesEncoder
 from meeko import PDBQTMolecule
 from meeko import RDKitMolCreate
 from meeko import MoleculePreparation
@@ -186,7 +184,6 @@ def get_args():
                                   "the specified cache file and updates may be made to the same file in a cumulative manner. " 
                               ), 
                               nargs = "?", 
-                              default = False, 
     )
     config_group.add_argument("--mk_config", help="[.json]", metavar="JSON_FILENAME")
     config_group.add_argument(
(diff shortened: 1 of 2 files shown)


rwxayheee (approved):
This is a nice working fix, I approve the changes. If possible, ignoring excess bonds or activating a switch of matching rules at certain atoms (specified by patterns or labels) might be safer. The multiple-round mapping with different forgiving rules is kind of doing  this

diogomart:
Thanks. The clashes can occur between any atoms, I don't think it's possible to restrict the forgiving of extra inter-residue bonds to certain atoms. Here's an Ile-Met clash.
trouble

rwxayheee:
Yes, users just need to be aware of what is being forgiven. To forgive this in 2nd and 3rd round means we're particularly strict with the 1st round (embedded): 

```
In the 1st round
residues that match exactly with the embedded templates (MET) will pass

2nd round (only happens when no match in 1st round)
residues that could be embedded will pass, when we forgive H_excess and bond_missing at linkers

3rd round (only happens when no match in 1st and 2nd rounds)
residues that are exactly the terminal (NMET, CMET) will pass
```

If such violation occurs at backbone, maybe we will have a tie into the second round (?) I don't have an example. Based on the existing tests, the changes work out well. I approve the change

rwxayheee:
In these structures - Can the templates be directly assigned, instead of getting from matching (it's just not worth it)? The matching is a bit too dynamic (I regret implementing it in this way) and it's a bit difficult to know how certain violation of the rules will change the flow of the templates. But again, at the moment I don't see an issue with the PR. This PR seems to be a working solution

diogomart:
Just added the warning with logging, it looks like this
```
matched with excess inter-residue bond(s): A:423
matched with excess inter-residue bond(s): A:424
```
Yeah I'm also not 100% sure this more forgiving approach will never fail. But if neither of us can think of an example to break it, it should be not too crazy.... Thanks for looking into it.

rwxayheee:
I think excess bonds at sulfur atom in CYS might lead to unexpected matching (but I'm not 100% sure). The matched template is technically correct for the given structure, but in fact it might not be a valid disulfide bond. In these cases, direct assignment might be easier than going through the template matching, if we know what templates to assign. 

For more complicate, malformed structures like these, the optimal correction happens in bonds (before template matching). Doing a pattern-based unbonding could help

PR #348, opened 2025-04-23
rigidify triple bonds (rigidify single bonds in single-triple-single)

### LINKED ISSUE #127: nitriles are rotatable
Not a problem, for example vina rotates hydroxyls while ignoring H and that's fine, but it's a tiny bit suboptimal and is now noted herein.

diogomart:
Fixes #127 

Triple bonds are rigid.

Single bonds in nitriles are also rigid.

Implemented @rwxayheee idea of, in single-triple-single bond systems, rigidifying the single bonds and rotating the triple bond instead, as this is the most efficient approach.

### DIFF
--- a/meeko/bondtyper.py
+++ b/meeko/bondtyper.py
@@ -36,6 +36,19 @@ class BondTyperLegacy:
                 num_amides_removed += 1
         assert num_amides_originally == num_amides_removed + len(amide_bonds)
 
+        triple_bonds = [(x[0], x[1]) for x in setup.find_pattern("[*]#[*]")]
+        single_triple_single = [
+            (x[0], x[1], x[2], x[3]) for x in setup.find_pattern("[*]-[*]#[*]-[*]")
+        ]
+        single_to_rigidify = []
+        for i, j, k, m in single_triple_single:
+            triple_bonds.remove((j, k))
+            single_to_rigidify.append((i, j))
+            single_to_rigidify.append((k, m))
+        # fully rigidify nitrile and alike
+        single_to_rigidify.extend(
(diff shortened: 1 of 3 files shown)


rwxayheee:
Thanks for implementing this, sorry this might not be a very good idea. I have two minor concerns: 

1- Degeneracy due to merging nonpolar hydrogens
I don't 100% understand if we're determining the rotatable/non-rotatable property in the all-atom scope, or after merging nonpolar hydrogens..  

2- `"[*]-[*]#[*]-[*X1]"` (for example, "C-C#C-Cl" or "C-C#C-H") behaves more like terminal triple bonds
Currently, they have rotatable triple bonds that won't make a distinct conformer. 

This said I'm totally fine either way, having the hacky rotatable triple bonds or doing the chemically meaningful frozen triple bonds. Thanks again for looking into this

rwxayheee (approved):
The current implementation is technically correct, I approve the change

diogomart:
> I don't 100% understand if we're determining the rotatable/non-rotatable property in the all-atom scope, or after merging nonpolar hydrogens

The code in `bondtyper.py` operates on the full atom scope, so a -CH3 with default AD4 types and non-polar H merging will be deemed rotatable. But then there's this code block in `preparation.py`:
https://github.com/forlilab/Meeko/blob/07a38f8bac4aa9eef4ef30aa8e67475b15d517aa/meeko/preparation.py#L277-L292

> 2- "[*]-[*]#[*]-[*X1]" (for example, "C-C#C-Cl" or "C-C#C-H") behaves more like terminal triple bonds
Currently, they have rotatable triple bonds that won't make a distinct conformer.

Yeah I gave up on those... 80% of the benefit with 20% of the effort :)

Thank you for reviewing!

PR #353, opened 2025-04-25
(Fix #352) Allow multiple matches from raw mol, when raw mol is symmetrical

### LINKED ISSUE #352: Receptor preparation fails for unexpected atom order
Here, CB and OG of residue 201 are misplaced. If they are moved to their correct place according to the serial number, this file works. The expected behavior is for meeko to work independently of atom order.

```
ATOM  1296  N    ASP A200       69.546  55.148  -0.977  1.00 19.96          N 
ATOM  1297  CA   ASP A200       68.531  54.617  -0.068  1.00 20.19          C 
ATOM  1298  C    ASP A200       67.830  53.395  -0.653  1.00 19.92          C 
ATOM  1299  O    ASP A200       67.569  52.420   0.053  1.00 20.88          O 
ATOM  1300  CB   ASP A200       67.501  55.696   0.285  1.00 22.41          C 
ATOM  1301  CG   ASP A200       68.091  56.814   1.132  1.00 23.03          C 
ATOM  1302  OD1  ASP A200       69.102  56.562   1.814  1.00 24.38          O 
ATOM  1303  OD2  ASP A200       67.529  57.932   1.129  1.00 28.47          O 1-
ATOM  1308  CB   SER A201       66.461  52.741  -4.028  0.50 22.04          C 
ATOM  1309  OG   SER A201       65.744  51.699  -4.664  0.50 24.33          O 
ATOM  1304  N    SER A201       67.531  53.444  -1.947  0.50 20.94          N 
ATOM  1305  CA   SER A201       66.865  52.335  -2.607  0.50 21.60          C 
ATOM  1306  C    SER A201       67.769  51.108  -2.647  0.50 23.02          C 
ATOM  1307  O    SER A201       67.325  49.983  -2.395  0.50 19.40          O 
ATOM  1310  N    TYR A202       69.046  51.315  -2.964  1.00 21.79          N 
ATOM  1311  CA   TYR A202       69.981  50.197  -3.006  1.00 21.61          C 
ATOM  1312  C    TYR A202       70.054  49.525  -1.639  1.00 19.61          C 
ATOM  1313  O    TYR A202       70.026  48.303  -1.534  1.00 19.95          O 
ATOM  1314  CB   TYR A202       71.393  50.665  -3.397  1.00 21.73          C 
ATOM  1315  CG   TYR A202       72.431  49.597  -3.139  1.00 20.55          C 
ATOM  1316  CD1  TYR A202       72.505  48.458  -3.943  1.00 17.40          C 
ATOM  1317  CD2  TYR A202       73.267  49.668  -2.019  1.00 18.03          C 
ATOM  1318  CE1  TYR A202       73.373  47.414  -3.638  1.00 20.50          C 
ATOM  1319  CE2  TYR A202       74.138  48.628  -1.705  1.00 18.78          C 
ATOM  1320  CZ   TYR A202       74.186  47.504  -2.513  1.00 17.89          C 
ATOM  1321  OH   TYR A202       75.011  46.452  -2.190  1.00 19.61          O 
```

rwxayheee:
This seems to work for the specific case in #352 , but requires multiple matches from function match. Maybe it changed more than it should, and make the template matching even more complicated, but it seems to be a working solution.

### DIFF
--- a/meeko/export_flexres.py
+++ b/meeko/export_flexres.py
@@ -133,6 +133,7 @@ def export_pdb_updated_flexres(polymer, pdbqt_mol):
                     key = polymer.monomers[res_id].residue_template_key
                     template = polymer.residue_chem_templates.residue_templates[key]
                     _, template_to_pdbqt = template.match(mol)
+                    template_to_pdbqt = template_to_pdbqt[0]
 
                 sidechain_positions = {}
                 molsetup_matched = set()
--- a/meeko/polymer.py
+++ b/meeko/polymer.py
@@ -16,6 +16,7 @@ from rdkit.Chem import rdFMCS
 from rdkit.Chem import rdChemReactions
 from rdkit.Chem import rdMolInterchange
 from rdkit.Geometry import Point3D
+from collections import Counter
 
 from .molsetup import RDKitMoleculeSetup
 from .molsetup import MoleculeSetup
@@ -211,9 +212,6 @@ def mapping_by_mcs(mol, ref):
     mcs_mol = Chem.MolFromSmarts(mcs_result.smartsString)
     mol_matches = mol.GetSubstructMatches(mcs_mol)
     ref_matches = ref.GetSubstructMatches(mcs_mol, uniquify=False)
-    if len(ref_matches) > 2: 
-        ref_matches = ref_matches[:2]
-
 
     atom_maps = []
     for mol_idxs in mol_matches:
@@ -221,7 +219,18 @@ def mapping_by_mcs(mol, ref):
             if len(mol_idxs) == len(ref_idxs):
                 atom_maps.append({i: j for i, j in zip(mol_idxs, ref_idxs)})
 
-    return atom_maps
+    def symmetry_score(mol):
+        """Returns a rough score of symmetry: higher = more symmetric."""
+        ranks = Chem.CanonicalRankAtoms(mol, breakTies=False)
+        rank_counts = Counter(ranks)
+        score = sum(count for count in rank_counts.values() if count > 1)
+        return score
+    
+
+    if symmetry_score(ref)>symmetry_score(mol): # ref is more symmetric
+        return atom_maps
+    else:
+        return [atom_maps[0]]  # return only the first one
 
 
 def _snap_to_int(value, tolerance=0.12):
@@ -1474,10 +1483,6 @@ class Polymer(BaseJSONParsable):
             "msg": "",
         }
         for residue_key, (raw_mol, input_resname) in raw_input_mols.items():
-            if input_resname in ["SER"]: 
-                match_multiple = True
-            else:
-                match_multiple = False
             if raw_mol is None:
                 monomers[residue_key] = Monomer(
                     None, None, None, input_resname, None
@@ -1542,7 +1547,7 @@ class Polymer(BaseJSONParsable):
             for index, template in enumerate(candidate_templates):
 
                 # match intra-residue graph
-                results, matched_mappings = template.match(raw_mol, multiple=match_multiple)
+                results, matched_mappings = template.match(raw_mol)
                 if len(matched_mappings) > 1: 
                     print(f"Warning: multiple matches found for"
                           f" {candidate_template_keys[index]}")
@@ -3080,7 +3085,7 @@ class ResidueTemplate(BaseJSONParsable):
             raise ValueError(f"{len(atom_names)=} differs from {mol.GetNumAtoms()=}")
         return
 
-    def match(self, input_mol, multiple=False):
+    def match(self, input_mol):
         mappings = mapping_by_mcs(self.mol, input_mol)
         results = []
         for mapping in mappings: 
@@ -3122,10 +3127,7 @@ class ResidueTemplate(BaseJSONParsable):
                         else:
                             result[element]["excess"].append(-1)
             results.append(result)
-        if multiple:
-            return results, mappings
-        else:
-            return [results[0]], [mappings[0]]
+        return results, mappings
 
 # region JSON Encoders
 
--- a/meeko/polymer.py
+++ b/meeko/polymer.py
@@ -204,13 +204,21 @@ def mapping_by_mcs(mol, ref):
 
     """
     mcs_result = rdFMCS.FindMCS([mol, ref], bondCompare=rdFMCS.BondCompare.CompareAny)
+
+    if not mcs_result.smartsString:
+        return []
+
     mcs_mol = Chem.MolFromSmarts(mcs_result.smartsString)
+    mol_matches = mol.GetSubstructMatches(mcs_mol, uniquify=False)
+    ref_matches = ref.GetSubstructMatches(mcs_mol, uniquify=False)
 
-    mol_idxs = mol.GetSubstructMatch(mcs_mol)
-    ref_idxs = ref.GetSubstructMatch(mcs_mol)
+    atom_maps = []
+    for mol_idxs in mol_matches:
+        for ref_idxs in ref_matches:
+            if len(mol_idxs) == len(ref_idxs):
+                atom_maps.append({i: j for i, j in zip(mol_idxs, ref_idxs)})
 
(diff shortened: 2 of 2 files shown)


rwxayheee:
#### Explanation of #352 
This is not a bonding issue, I was thinking about something else that is not relevant. This special case was failing because in the past, only one (arbitrary) match is considered from function `mapping_by_mcs` and `match`. 

For unprotonated Serine, at the point of mapping evaluation against the template: `mapping_by_mcs(self.mol, input_mol)`
`self.mol` is a `template.mol`, which is asymmetric: `(Smiles) [H]NC([H])(C=O)C([H])([H])O[H]`
`input_mol ` is the `raw_mol`, which is symmetric: `(Smiles) [N][C]([C][O])[C][O]`

Therefore, two matches of the MCS in `raw_mol` need to be considered, as there's a vanishing of symmetry when the template is enforced on raw_mol. 

#### Solution
The vanishing of symmetry is evaluated by an empirical symmetry score, not sure if it works in all cases: 

```
    def symmetry_score(mol):
        """Returns a rough score of symmetry: higher = more symmetric."""
        ranks = Chem.CanonicalRankAtoms(mol, breakTies=False)
        rank_counts = Counter(ranks)
        score = sum(count for count in rank_counts.values() if count > 1)
        return score
```

And because the return type of `match` and `mapping_by_mcs` have become lists, and one template can possibly have multiple mappings being evaluated, some minor adjustment is made (indexing, etc.)

diogomart:
Wow, this bug was really deep. I did not expect that. Thank you so much.

diogomart:
The symmetry_score depends on whether the molecule has hydrogens or not:

```python
from rdkit import Chem
from collections import Counter

def symmetry_score(mol):
    """Returns a rough score of symmetry: higher = more symmetric."""
    ranks = Chem.CanonicalRankAtoms(mol, breakTies=False)
    rank_counts = Counter(ranks)
    score = sum(count for count in rank_counts.values() if count > 1)
    return score

p = Chem.SmilesParserParams()
p.removeHs=False
asp_template = Chem.MolFromSmiles("C(=O)C([H])(N[H])C([H])([H])C(=O)[O-]", p)
asp_template_noH = Chem.MolFromSmiles("C(=O)C([H])(N[H])C([H])([H])C(=O)[O-]")
asp_raw = Chem.MolFromSmiles("C(O)C(N)CC(O)O")

ser_template = Chem.MolFromSmiles("C(=O)C([H])(N[H])C([H])([H])O[H]", p)
ser_template_noH = Chem.MolFromSmiles("C(=O)C([H])(N[H])C([H])([H])O[H]")
ser_raw = Chem.MolFromSmiles("C(O)C(N)CO")


print(f"{symmetry_score(ser_template)=}")
print(f"{symmetry_score(ser_template_noH)=}")
print(f"{symmetry_score(ser_raw)=}")
print()
print(f"{symmetry_score(asp_template)=}")
print(f"{symmetry_score(asp_template_noH)=}")
print(f"{symmetry_score(asp_raw)=}")
```

Here's the result:
```
symmetry_score(ser_template)=2
symmetry_score(ser_template_noH)=0
symmetry_score(ser_raw)=4

symmetry_score(asp_template)=2
symmetry_score(asp_template_noH)=0
symmetry_score(asp_raw)=2
```

Thus, I think we need an alternate solution, as the raw mols may or may not have hydrogens. What about keeping multiple mapping only when the indices of link atoms are affected?

rwxayheee:
> Thus, I think we need an alternate solution, as the raw mols may or may not have hydrogens. What about keeping multiple mapping only when the indices of link atoms are affected?

We can exclude hydrogens from the symmetry evaluation. (But I think usually template has hydrogens, while raw mol is typically missing hydrogens so it shouldn’t matter)

I have thought about using link atoms too. but the information is not passed on the layer. If you have a suggestion, please make the change and I can take a second look

diogomart:
For templates and raw molecules that match, templates without hydrogens have a smaller symmetry score because they have bond orders, while all bonds are single in raw molecules.

I'll look into implementing the link atoms and PR to your PR.

rwxayheee:
> For templates and raw molecules that match, templates without hydrogens have a smaller symmetry score because they have bond orders, while all bonds are single in raw molecules

Yes, this is why raw mol is more symmetrical than template and therefore multiple matches are needed. So I think the symmetry score should work to pick out symmetrical raw mol like unprotected serine. Maybe I still don’t understand the templates without hydrogen scenario you described.. 

I think the match needs to account for vanishing symmetry in this way, regardless of link atoms or not. But doing this can generate multiple protonated conformers and it’s not the current behavior (just construct an arbitrary protomer). To keep things simple, we can use link atoms for now. If people raise other issues about the template matching in the future, a redesgn would be helpful.

diogomart:
I think that without hydrogens this statement is never true, but it's OK to always return all the mappings <- EDITED
https://github.com/rwxayheee/Meeko/blob/569654fe91957c4b882f8e66788596d21df6996f/meeko/polymer.py#L230

> I think the match needs to account for vanishing symmetry in this way

Could you clarify what you mean by "vanishing symmetry"? To clarify, I think the best is to return all the matches accounting for symmetry. I really like your implementation.

rwxayheee:
> Could you clarify what you mean by "vanishing symmetry"? To clarify, I think the best is to return all the matches accounting for symmetry. I really like your implementation.

I commented in the code (but maybe it's bad and confusing language..). Here's an explanation: 

The match function is a function that tries to apply a template onto a raw mol. The vanishing of symmetry (raw -> template) means there are multiple equally valid mappings between raw mol and template. 

The raw mol in the specific case (unprotonated Serine) is lacking hydrogens and bond order, and therefore it becomes highly symmetrical. As you pointed out as an alternate solution, if taking into account the link atoms, this raw will not be symmetrical. But currently, the link atoms evaluation happens after the matching. Therefore, the solution in is PR is to return multiple matches for further evaluation. 

A potential problem is that this PR might produce some tied mappings that need manual selection & inspection. (This is actually needed, for a single, unprotonated Serine as there's no way to distinguish C(=O) and CB-OH?) So in theory, we should generate all possible outcomes for these but this might be out of the scope of Meeko at this moment, as we always recommend protonating receptors before making the PDBQT. 

Both #347 and #352 are applied cases that are kind of outside the original design scope. I understand the urgency but if there's no optimal solution, please don't rush to solve it. I recommend fixing the PDB files in applied projects, if possible.

diogomart:
I think this is good to merge after https://github.com/rwxayheee/Meeko/pull/8/files

diogomart:
> A potential problem is that this PR might produce some tied mappings that need manual selection & inspection. (This is actually needed, for a single, unprotonated Serine as there's no way to distinguish C(=O) and CB-OH?) So in theory, we should generate all possible outcomes for these but this might be out of the scope of Meeko at this moment, as we always recommend protonating receptors before making the PDBQT.

I think this is addressed by this block, but let me know if you have concerns:
https://github.com/diogomart/Meeko/blob/b129b22f17a43f0c292a807de1d41051139a4f1e/meeko/polymer.py#L1625-L1636

rwxayheee:
Hi @diogomart 

Returning all matching and unifying by template names is not what I intended. 

Returning all matching: 
This can return too many mappings

Unifying by template names: 
This may not be correct. Same template can have distinguishable mappings. 

If you're looking for a quick fix for this issue, I will approve the changes. But the changes are very different from what I was planning. If we're not account for the symmetry, this PR should be closed and a different approach should be taken.

rwxayheee:
```
                # TODO 
                # We need mappings to cover different indices of link atoms only.
                # Carboxylates (e.g. ASP, GLU) will have two mappings for the
                # carboxylate oxygens, but we don't care about this and could filter
                # these out. We care about residues like serine, in which the
                # hydroxyl and carbonyl are equivalent in the raw mol because all
                # bonds are single.

```

If unprotonated ASP/GLU is mapping to ASH/GLH, there's a vanishing symmetry and two possible mappings under the same template key. 

Maybe it's not worth too much work at this time, to account for exceptions like unprotonated Serine or clashes in sidechains. The matching logic is too complicated and I feel bad for making it more complicated..

Issue #192, opened 2024-10-12
Include screening examples in doc


rwxayheee:
Hi, I just pushed the Colab examples to readthedoc. 
https://meeko.readthedocs.io/en/readthedocs/colab_examples.html

There are currently 6 examples (4 Vina + 2 AD-GPU). Each is one single-docking experiment. The procedure is slightly different but the general workflow is the same. 

- [x] These examples may need some final touch to be up to date with the recent changes we made to the command line scripts. 

Specifically: 
- Syntax change & a brief explanation (-o <basename> -v -g -j) 
- Branch: use release when we have the next one

The updates will be made directly on the linked notebook. Therefore, it's safe to merge before we finalize the notebook contents. 

- [x] I'm adding 1-2 more screening examples in future. 

I have a plan but I haven't started writing them yet, so I don't have an ETA... They will use AD-GPU and involve post-processing by Ringtail. And a different workflow picture.

rwxayheee:
Merged [#354](https://github.com/ccsb-scripps/AutoDock-Vina/pull/354). The existing notebooks and published tutorials for Vina & meek are up to date with the current release (v0.6.0). Will start task 2 after some revision of the rsd3 website

rwxayheee:
Will be adding two new tutorials to the doc: 

Tutorial 4A - Preparation and execution of virtual screening

Tutorial 4B - In-depth analysis: Basic AUC-ROC/EF assessments, and building & interpretation of advanced regression models (XGBoost for example) involving interaction vectors 

These will be based on a small virtual screening (single receptor, about 1000 ligands = 500 actives + 500 decoys) which should be doable on a regular laptop using AutoDock-GPU. It will be based on an actual protein and an actual ligand data set, but the conclusion should not be judged professionally. Hoping these examples provide useful code snippets such as how to reconstruct molecules and how to vectorize the interactions stored in the database

rwxayheee:
Closing as completed

PR #359, opened 2025-05-07
Merge manual doc updates before rebasing release-doc

rwxayheee:
This PR will merge the two examples mentioned (#192) (and some other earlier changes) that were added directly to release-doc, which are applicable to both the current release and the upcoming version. There are no visible conflicts based on current branch state. 
No immediate action is required, but the manually changed content like this in release-doc might need to be merged before release-doc is rebased.


joanimato:
This will be done with a recent PR, so probably no longer needed.

PR #361, opened 2025-05-08
only single bonds are rotatable (mostly)

diogomart:
fixes bug introduced in c3550596a3a2b9003abd279f11eee8499d58beba


rwxayheee:
This looks good, do you find a conflict with #348 (wants rotatable triple bond) or we're ok with the current behavior (this PR)? 
Can you provide some more description of what the bug was, and what is the expected behavior after this PR? I don't fully understand the usage of rotatable bonds, so some background would be really helpful

diogomart:
The bug caused double bonds to rotate as in the figure below.

I think you are right: it clashes with #348, I'll add another test and fix if bug/clash confirmed.

alkene

joanimato (approved):
Looks good to me.

PR #366, opened 2025-06-03
support altloc B in PDB for mk_prepare_receptor.py --box_enveloping

diogomart:
The RDKit PDB parser ignores atoms with altloc "B" (or any altloc that isn't either "A" or empty " "). This PR pre-processes the PDB string to set the altloc of all atoms to an empty string. This is exclusively for PDB files passed to `mk_prepare_receptor.py --box_enveloping` argument.


nbruciaferri (approved):
Looks good to me!

PR #375, opened 2025-07-30
bugfix dihedrals overcounted by non-canonicalized atom indices

diogomart:
The fix canonicalizes atom indices defining dihedral paths making it impossible for the same set of four atoms `(a, b, c, d)` and `(d, c, b, a)` to define two dihedral potentials.


diogomart:
The data structure to store the parameters uses the indexes as a key in a python dictionary, so we have
```
molsetup.dihedral_partaking_atoms[(1, 2, 3, 4)] = some_index
```
and before the bug fix we had

```
molsetup.dihedral_partaking_atoms[(1, 2, 3, 4)] = some_index
molsetup.dihedral_partaking_atoms[(4, 3, 2, 1)] = another_index
```

meaning that one path of four atoms was assigning two different (!) dihedrals when it can only assing one. This happens because the SMARTS pattern go from more general to more specific. Here's a made up example:

```
[*:1][C:2][C:3][*:4] -> general parameter for C-C bonds
[*O:1]=[C:2][C:3][*:4] -> acetone-like
```

All patterns are matched in order. Atoms 1,2,3,4 will get the general parameter from the first iteration, and this will be replaced by the acetone parameter in the second iteration. But if when matching the acetone parameter we get 4,3,2,1, then we keep both parameters when we should be replacing the first.

`some_index` and `another_index` point to `molsetup.dihedral_interactions` where all parameters are stored (k, periodicity, and phase) for the full set of periodicities

meeko/atomtyper.py
atillack, line 1: Just a recap before someone in the future thinks "this canonicalization seems unnecessary": The problem is that multiple SMARTS patterns may match the same dihedral yet differing by their indices' direction - so the canonicalization makes sure these cases don't lead to an additional dihedral.

atillack (approved):

diogomart:
Thank you @atillack for reviewing and for finding the bug!

PR #377, opened 2025-08-06
assign espaloma charges to single atom molecules

diogomart:
Assigning charges to ions caused trouble because it's impossible to create an espaloma graph with just one atom. The error was thrown by espaloma. In this PR, if the molecule has one atom, its espaloma charge will be the formal charge.

Also moved two methods specific to Gasteiger charges from `RDKitMoleculeSetup` to `rdkitutils.py`.


joanimato (approved):

mallanos:
Looks good to me, thanks for taking care!

PR #386, opened 2025-08-19
Gasteiger charges working for Iron-Sulfur clusters

diogomart:
The RDKit implementation of Gasteiger charges does not have parameters for Iron, so Iron is removed, hydrogens are added (in a temporary molecule) to satisfy the valence left open by Iron, charges are computed on the temporary molecule, and then the charge of the added hydrogens is summed onto the Iron atom in the real molecule (in addition to its formal charge). This was working for heme. This PR enables the same idea to work for Fe-S clusters, by allowing more than one H addition per atom in the temporary molecule (in Fe-S clusters some Sulfurs will get two atoms). Also, some Sulfurs in the temporary molecule will donate the charge of their added hydrogens to more than one Iron, in which case this PR divides the donated charge by the number of donations, to avoid double counting.

### DIFF
--- a/meeko/utils/rdkitutils.py
+++ b/meeko/utils/rdkitutils.py
@@ -496,29 +496,34 @@ def compute_gasteiger_charges(rdkit_mol):
     charges = [a.GetDoubleProp("_GasteigerCharge") for a in copy_mol.GetAtoms()]
     if idx_rm_to_formal_charge:
         ok_charges = charges.copy()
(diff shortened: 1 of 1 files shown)


allisonbarkdull (approved):
looks good; I tested this for the system I showed you earlier and it ran as expected.
My only thought is that it may be useful to print a warning when there are no gasteiger charges for a given element so the user knows what's going on.

PR #411, opened 2025-10-30
feat: add JSON molsetup to RDKit mol prop for SDF storage

diogomart:


diogomart:
Not ready yet: need to add functionality to mk_prepare_ligand.py to write SDF as output.

sforli:
What other software needs to be updated to be able to parse these SDF?

diogomart:
These are readable by other software, e.g. Pymol. The JSON string in the property is useful only for Meeko, no other software needs to use it. The size increase is about 5x after compression, I now think this is a bad idea because of how much space it takes, and that we should store only what is expensive to compute, which are the partial charges.

atillack:
I agree with Diogo, everything but the charges in the molsetup can be quickly recreated and thus doesn't seem worth the cost in terms of space and/or unzipping time.

sforli:
I meant our own software (including Meeko itself) to be able to read it back and benefit from having cached properties.

OK to drop it for non-charge properties

diogomart:
`charge_model="read"` reads charges from property fields, see in [preparation.py](https://github.com/forlilab/Meeko/blob/3c6795f6a3c6402524782674a87c877797466c7c/meeko/preparation.py#L165) and in [molsetup.py](https://github.com/forlilab/Meeko/blob/3c6795f6a3c6402524782674a87c877797466c7c/meeko/molsetup.py#L1646)

We still need to implement writing the charges to RDKit properties

As for this PR, I'll close it, as we concluded the disk space costs do not justify the benefit.

PR #412, opened 2025-10-30
nagl charges from latest version of openff

joanimato:
Implements NAGL charges here: https://docs.openforcefield.org/projects/nagl/en/latest/theory.html

Requires a fairly recent version of openff dependencies. Not sure what the best way is to enforce that.

### DIFF
--- a/meeko/cli/mk_prepare_ligand.py
+++ b/meeko/cli/mk_prepare_ligand.py
@@ -206,7 +206,7 @@ def cmd_lineparser():
     )
     config_group.add_argument(
         "--charge_model",
-        choices=("gasteiger", "espaloma", "zero", "read"),
+        choices=("gasteiger", "espaloma", "zero", "read", "nagl"),
         help="default is 'gasteiger', 'zero' sets all zeros",
         default="gasteiger",
     )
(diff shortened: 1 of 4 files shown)


diogomart:
I'd move the `openff.toolkit` import inside the NAGL charge block because
 - it takes a couple seconds to import
 - make the dependency optional
 
 and then you can break down  the `except` block into two:
  - `except ImportError` with a message about a missing dependency
  - a broader `except Exception` to catch other possible NAGL errors
  
 I'd always print the exception (as this PR is already doing) with `except Exception as err; print(err)` because the detailed error can help troubleshoot.

diogomart:
The first import takes 3s, calling it again just a few microseconds (probably Python checking it's already imported.
```python
>>> from time import time
>>> for i in range(5):
...     t0 = time()
...     from openff.toolkit import Molecule
...     print(f"{time() - t0:.7f}")
... 
3.1275003
0.0000050
0.0000031
0.0000024
0.0000031
```

joanimato:
> I'd move the `openff.toolkit` import inside the NAGL charge block because
> 
> * it takes a couple seconds to import
> * make the dependency optional
> 
> and then you can break down the `except` block into two:
> 
> * `except ImportError` with a message about a missing dependency
> * a broader `except Exception` to catch other possible NAGL errors
> 
> I'd always print the exception (as this PR is already doing) with `except Exception as err; print(err)` because the detailed error can help troubleshoot.

Updated.

diogomart (approved):

PR #418, opened 2025-11-13
feat: arg only_hs_with_coords exports mols with only Hs present in docking

diogomart:
Dockings with Vina and AD4 scoring functions do not use non-polar Hs (atom type H). When exporting poses with `mk_export.py` or `RDKitMolCreate`, RDKit adds non-polar hydrogens. The new option (which is not enabled by default) exports poses without Hs that were absent in docking. This PR also deletes an obsolete function that is never used.


maylinnp (approved):
looks good to me!

PR #419, opened 2025-11-14
feat: optionally move vdW volume when merging terminal atoms

diogomart:


atillack:
@diogomart Awesome 😎 Works for me - we could save a few clock cycles by directly calculating the new radius as `r_new = cbrt(r_orig^3 + r_merged^3)` but i don't think it'll have a measurable impact 😅

atillack (approved):

PR #421, opened 2025-11-17
Fix: rotatable terminal groups use user-defined mk_config

diogomart:


joanimato (approved):
Tested it, everything seems to work now. The parameters are generated in the json file and zeta runs correctly.

PR #424, opened 2025-12-04
fix: do not uniquify SMARTS matches by default (for OpenFF compat)

diogomart:
The bug was discovered by inspecting water parameters, each hydrogen had different parameters. One H had TIP3P parameters, while the other had hydroxyl parameters. This happened because the OpenFF XML forcefield file specifies both hydrogens in the SMARTS, for example `[H:1]O[H]` but only the one with the ":1" label is parameterized. With `uniquify=True` (RDKit's default, in use before this PR) only one match is returned. By setting `uniquify=False`, RDKit returns two matches instead of one, in which each match labels a different H.

### DIFF
--- a/meeko/atomtyper.py
+++ b/meeko/atomtyper.py
@@ -99,7 +99,7 @@ class AtomTyper:
             for line in offatom_params[smartsgroup]:
                 # SMARTS
                 smarts = str(line["smarts"])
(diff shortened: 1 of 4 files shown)


allisonbarkdull (approved):
lgtm

PR #425, opened 2025-12-05
perf: vina_params merges all H

diogomart:


diogomart:
By default, atoms typed "H" are merged, setting `is_ignore=True` in the molsetup. This will exclude them from calculation, likely leading to a speed up.

diogomart:
Note: not expecting a speedup with standard Vina using PDBQT input, as the PDBQT already excludes non-polar H (type "H"), and keeps polar H (type "HD") for typing at C++ level by Vina program

PR #428, opened 2025-12-10
feat: inflexibilize polymer sidechains

diogomart:


meeko/polymer.py
```diff
@@ -2609,6 +2621,12 @@ def flexibilize(self, mk_prep):
                     self.is_flexres_atom[atom_idx] = True
         return
 
+    def inflexibilize(self, mk_prep, residue_id):
```
joanimato: Should these functions be named "rigidify"?

PR #431, opened 2026-01-05
json receptors inputs for mk_prepare_receptor.py

joanimato:
This PR adds additional option to input json receptors into `mk_prepare_receptor.py`. The receptor is prepared and parametrized just as the pdb option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Meeko is developed by the [Forli lab](https://forlilab.org/) at the
 [Center for Computational Structural Biology (CCSB)](https://ccsb.scripps.edu)
 at [Scripps Research](https://www.scripps.edu/).
 
-If you use this, please cite: ["Meeko: molecule parameterization and software interoperability for docking and beyond"](https://chemrxiv.org/engage/chemrxiv/article-details/68c054243e708a7649fa21d3)
+If you use this, please cite: ["Meeko: molecule parameterization and software interoperability for docking and beyond"](https://pubs.acs.org/doi/full/10.1021/acs.jcim.5c02271)
 
 
 ## Documentation
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@ Meeko is developed by the [Forli lab](https://forlilab.org/) at the
 [Center for Computational Structural Biology (CCSB)](https://ccsb.scripps.edu)
 at [Scripps Research](https://www.scripps.edu/).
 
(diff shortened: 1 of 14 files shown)


diogomart (approved):
Looks good to me, added a warning that existing parameters and flexres are lost. A user could have save a JSON with a specific charge model (or other params) and run mk_prepare_receptor to write a PDBQT, and assume that the charges stored in the JSON would be preserved. Feel free to merge.

joanimato:
I think it might be a good idea to preserve the charges if `--charge_model` is not supplied by the user, since the charges are also the most time consuming to compute. I'll see if I can add that. If not, then it can be a future PR.

diogomart:
What use cases are you thinking for `--read_json`? I have these two in mind:
 - write a PDBQT from existing JSON
 - read JSON, edit flexible sidechains (add/remove/terminal), and save JSON

Both would be better served by keeping the existing parameterization. But then passing options that change the parameterization, such as `--charge_model` should raise an error or trigger re-parameterization. One difficulty is that it's not possible to keep some parameters and edit others.

joanimato:
Those are two common scenarios, but I also had in mind the case where someone wants to use meeko as they normally would and only have the json available (or the json is more convenient). In my opinion, re-parametrization should be the default because that's what someone would expect when running the CLI. With the API it's a different story. 

Moreover, I don't know if it's immediately obvious to know how a given json receptor was prepared. For instance, I don't think the json file contains the identity of the charge model used to prepare the receptor, so unless labeled in some other way, a user would not know.

diogomart:
Sounds good to re-parameterize by default. There's no way to know how a polymer was prepared, because parameterization choices (such as charge model) are not stored in the Polymer.

joanimato:
Now there is an option to use charge_model = json, which will store the charges from the json input and pass them through to the polymer/monomer objects. The rest of the pipeline remains the same.

diogomart:
I realized we can't keep charges from JSON. It's possible to change the protonation state of some residues with argument `--set_template`, e.g. change a hystidine from HIE to HID, or even remove hydrogens from a Lys by setting the template with LYN, and in these cases charges need to be re-calculated.

diogomart:
I'll merge 852d2d1 and close this PR

diogomart:
ah, can't push directly. Shall we create a new PR from that commit?

joanimato:
Let me see if I can revert to that commit.

PR #434, opened 2026-01-09
Feat: Add option to ignore https certificate for PDB download

atillack:
Analogous to the option `--no-check-certificate` in wget, this PR adds the option `--ignore_https_cert` (default is False) to bypass https certificate checks for downloads from the PDB database as a last resort in systems with either misconfigured certificate providers or website misconfiguration.

This is a somewhat dangerous thing to do as spoofing of rcsb.org would not be detected, thus plenty of warning notes are included.


diogomart:
Awesome, thank you

diogomart (approved):

PR #435, opened 2026-01-21
Precomputed charges

joanimato:
~Not finished, still needs testing.~

This PR allows the option to skip charge computation and read from a template of precomputed charges stored in `template_charges.json`.


joanimato:
I did some tests comparing jsons with pre-computed and re-computed charges and everything seems to match. Everything should work now. There was some confusion because the charges stored in the polymer object when using the python API don't match what is printed in the json. I'm not sure if that's a bug, or if I'm missing something.

diogomart (approved):
Awesome, so much faster now

PR #442, opened 2026-01-26
feat: add pytest workflow

joanimato:
Runs pytest on every new PR. To be merged after #441


diogomart:
We had previously excluded requirements at the pip level to allow the user to manage dependencies manually. This was necessary to install meeko on Python>=3.12 because Prody did not have conda recipe for >=3.12. But maybe none of this matters because pip can be run with `--no-deps` option.

joanimato:
This should not affect the pypi workflow since the requirements.txt file is not invoked there. Still, we can remove the file and just install the requirements for this test. They have to be installed though, otherwise the tests won't run.

diogomart (approved):

diogomart:
The `requirements.txt` didn't trigger the installation of dependencies in a local `pip install -e .`

PR #443, opened 2026-01-27
feat: speed up (and rename) polymer.stitch function.

joanimato:
This renames the Polymer.stitch function to `to_rdkit_mol` though an alias `stitch` function is also created for backward compatibility. 

The function is modified for some significant speed up,

1. The function `find_inter_mols_bonds` is not called if it is not needed, since that was already called in the polymer creation. Instead, a class variable is defined that contains the bonds. 
2. The combination of the mols is modified to happen in NlogN time rather than N^2 time. This achieves a significant speed up.


joanimato:
It seems one test fails:

FAILED test/polymer_creation_test.py::test_stitch_polymer - assert False

I will investigate to see what's causing this.

joanimato:
I guess the self.bonds variable was not being written to json, so it was empty when read from json. Perhaps that can be addressed in a future PR if we want that quantity to be part of the json file.

diogomart (approved):

PR #449, opened 2026-02-10
fix: enable specifying openff version

diogomart:
Enable specifying OpenFF version for both vdW and torsion parameters. The string "openff" still specifies openff-2.0.0 for backwards compatibility, but explicit versions are now supported, e.g. "openff-2.3.0".

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ ] If applicable, documentation was updated with new feature. 

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,5 @@
+Dependencies for building locally with `make html`
+```
+micromamba install sphinx
+pip install sphinx-design sphinx-book-theme
+```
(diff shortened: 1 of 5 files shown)


joanimato (approved):
looks good

joanimato:
Does this need some documentation? Currently I don't see anything in the docs about what can go on the config json (which I don't think it's 100% the same as the cli options). 

Also, not related to this PR, but at some point we should upload the API docs.

diogomart:
Added info to docstring. Will consider adding to reathedocs.

diogomart:
Added detailed documentation to `Ligand preparation -> advanced usage`

PR #450, opened 2026-02-19
chore: remove prints and progress bar from polymer

diogomart:
## Description

Remove print statements and progress bar from `Polymer.parameterize()` function. These are very useful when running `mk_prepare_receptor.py` for a single protein, but clutter the stdout when running things in batch. Also, print statements when using the Python bindings directly can be annoying.

Added classmethod `Polymer.from_pdb_file` to spare the user from loading file contents into a string. These method simply wraps the existing `Polymer.from_pdb_string`. Docs added.

## Type of change

- [x] Documentation update
- [x] Other (please describe):

## Checklist:

- [ ] (Optional) new test added to account for new feature.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -102,6 +102,7 @@ to run molecular docking and virtual screening.
    Overview <rec_overview>
    Info on templates  <py_build_temp>
    Command line usage <cli_rec_prep>
+   Python bindings <py_rec_prep>
    mk_prepare_receptor.py options <rec_cli_options>
 
 .. toctree::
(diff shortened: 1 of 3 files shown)


joanimato (approved):
:(

diogomart:
Actually, we could add a print_stuff argument, which defaults to False, but mk_prepare_receptor.py sets it to True. Would that be good?

joanimato:
Yeah, that's a good idea, there could be a `verbose` option. But we can worry about printing in another PR, since we have to think about what info would be useful to print out.

PR #451, opened 2026-02-20
feat: package scofu1 config for mk_prepare_receptor.py --mk_config

diogomart:
## Description

This PR packages the configuration for a new scoring function (scofu1) in a JSON file in `data/mk_configs/`.

## Type of change

## Checklist:

- [ ] (Optional) new test added to account for new feature.

### DIFF
--- a/docs/source/rec_cli_options.rst
+++ b/docs/source/rec_cli_options.rst
@@ -72,8 +72,7 @@ Receptor Perception Options
    cumulative manner.
 
 .. option:: --mk_config
-   Can be a local json config file instead, or a the unsuffixed filename of a JSON file packaged
-   in meeko/data/mk_configs, for example --mk_config scofu1
+   local json config file [.json] or an unsuffixed (no .json) packaged file (choices: scofu1)
 
 .. option:: -d, --delete_bad_res
 
--- a/docs/source/rec_cli_options.rst
+++ b/docs/source/rec_cli_options.rst
@@ -71,8 +71,13 @@ Receptor Perception Options
    for polymer construction will be read from the specified cache file and updates may be made to the same file in a
    cumulative manner.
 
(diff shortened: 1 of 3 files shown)


docs/source/rec_cli_options.rst
atillack, line 1: Option text needs fixing, maybe to: "... local json config file or the unsuffixed ..."

diogomart: Good point, fixed

joanimato:
This might come down to personal preference, but I think the option `or an unsuffixed (no .json) packaged file (choices: scofu1)` is confusing. Since the mk_config file is just a way to for the user to bypass the cli options, does it make sense to have a default or presets? 

Also, meeko doesn't determine or decided the scoring function. In principle you can still run the output of this with another scoring function, correct?  If so, naming the config after the scoring function may be confusing for the user.

atillack:
i think having curated and more easily accessible presets for common scenarios is more user-friendly then letting users hunt for json files

allisonbarkdull:
I agree with both Andreas and Joani. 

I think the compromise is to have two separate arguments: maybe —soring_function scofu01 OR —custom_config config.json

(Not sure if —scoring_function makes the most sense semantically but you get the idea). I DEFINITELY think the config for our scoring functions need to be baked in, I do not want or keep track of that myself

diogomart:
I agree - @allisonbarkdull's suggestion is in agreement with Stefano's dislike of multiple uses for a single option. What about `--config_preset scofu1` and `--config_file my_local_file.json` ?

allisonbarkdull:
I like.

diogomart:
At least one approval is required before merging because the develop branch is protected

allisonbarkdull (approved):
lgtm

PR #467, opened 2026-05-21
Claude's refactor

diogomart:
This is a refactor by Claude to improve the code in general.

Lot's of stuff here, overall looks good. Tests pass and at least 1 test system with external code runs. Will keep this open while testing more integration with external programs. And also to see if we actually like the new structure better.


diogomart:
Both `PrepConfig` and `MoleculePreparation` define defaults. These should be defined in one place. It's also a code repetition. I think the idea of a `PrepConfig` dataclass is good but poorly implemented as is.

diogomart:
While implementing a feature on `merge_terminal_atoms` I realized Claude duplicated the function. It is now defined in two places:
```
meeko/molsetup/setup.py
meeko/molsetup/editing.py
```
Not good.

PR #481, opened 2026-07-02
fix: Update add_rotamers_to_polymer_molsetup to work with current Monomers.

atillack:
This PR fixes `add_rotamers_to_polymer_molsetup` in `polymer.py` to accept the current Monomer id's not containing the residue name in the key.


diogomart (approved):
Thanks!

PR #495, opened 2026-08-30
fix: define standard residue connectivity before sanitisation

ajrcampbell:
## Description

RDKit `DetermineConnectivity()` can add or omit bonds when PDB coordinates are distorted, causing standard residues to fail sanitisation before Meeko can match its bundled templates.

For complete named standard amino acids, this change builds the intramolecular graph from the single named-atom connectivity shared by applicable Meeko templates. It falls back to coordinate perception when no unique template graph applies. Nonstandard residues and inter-residue bonds are unchanged.

The regression fixture covers extra heavy-atom bonds, a missing heavy-atom bond, and an explicit hydrogen assigned to the wrong neighbor.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update
- [ ] Other (please describe):

## Checklist:

- [x] If applicable, documentation was updated with new feature. (Not applicable: no public API change.)
- [x] Docstring included for any new classes/functions/methods, or for significantly modified existing functions.
- [x] New test added to account for the bug fix.


meeko/utils/rdkitutils.py
```diff
@@ -379,7 +379,70 @@ def build_one_rdkit_mol_per_altloc(atom_fields_list):
...
+            )
+        )
+    if len(graphs) != 1:
+        return None
+
+    editable = Chem.RWMol(pdbmol)
+    for atom in editable.GetAtoms():
+        atom.SetNoImplicit(True)
+    indices = {name: index for index, name in enumerate(atom_names)}
+    for first, second in next(iter(graphs)):
```
ajrcampbell: The template path can reassign input coordinates to different atom names and produces hash-seed-dependent outputs.

I converted the four new fixtures with `PYTHONHASHSEED=1,2,3,4,5`. Every fixture produced different prepared PDB and PDBQT hashes across seeds. On `4ihl`, seed 2 swaps `NH1` and `NH2` and their hydrogen groups. The unordered graph at this loop changes bond insertion order, and the later name-blind MCS chooses a different symmetry mapping. The Synga test misses this because it compares names and coordinates as independent sets.

Please carry the validated input-name to template-name mapping into coordinate transfer. A regression should compare `(residue, atom name, element, coordinates)` and assert byte-identical prepared output across hash seeds. Sorting this loop alone would only make one arbitrary reassignment repeatable.

meeko/polymer.py
```diff
@@ -2530,11 +2557,20 @@ def _pdb_to_residue_mols(
...
+                    chem_templates.residue_templates[key]
+                    for key in template_keys
+                    if key in chem_templates.residue_templates
+                )
             try:
                 pdbmol, _, missed_altloc, needed_altloc = _aux_altloc_mol_build(
                     atom_field_list,
                     requested_altloc,
                     default_altloc,
+                    residue_templates,
```
ajrcampbell: This path still raises for any residue that falls back to coordinate perception and fails sanitisation, so one malformed residue aborts the complete receptor before `allow_bad_res` can apply.

I appended one invalid residue to a valid prepared receptor. The reviewed branch raises from `_pdb_to_residue_mols`. A scratch change that returns `pdbmol=None` for that residue produced usable output, reported only the added residue as unparsed and passed all 103 Meeko tests. The current 223-receptor snapshot does not contain this edge case, but the deleted Synga regression protected it.

Please return an unparsed result for the failing residue and let `handle_parsing_situations` enforce `allow_bad_res`, without restoring Synga's coordinate-only pre-pass. A mixed valid plus invalid receptor test should retain the valid residue and report the exact invalid residue.

meeko/polymer.py
```diff
@@ -1654,6 +1679,7 @@ def from_pdb_string(
             pdb_string,
             wanted_altloc,
             default_altloc,
+            chem_templates,
```
ajrcampbell: `set_template` is applied after this preprocessing step, so an explicit template cannot provide the graph used to sanitise distorted input.

With the distorted `5hxb` residue and `set_template={"A:577": "LYS"}`, default ambiguity succeeds. Changing `chem_templates.ambiguous["LYS"]` to `["LYN"]`, or naming the input `LYN`, raises before the explicit `LYS` selection is consulted.

Please pass the per-residue selection into `_pdb_to_residue_mols` and use it as the applicable template before sanitisation. Tests should cover both custom ambiguity and an explicit-state input name while preserving the selected template's named coordinates.

test/polymer_creation_test.py
```diff
@@ -51,6 +52,75 @@ def check_charge(residue, expected_charge, tolerance=0.002):
...
+            "NH2": ("HH21", "HH22"),
+        }
+    )
+    with open(distorted_standard_residues) as f:
+        raw_mols = Polymer._pdb_to_residue_mols(
+            f.read(),
+            chem_templates=chem_templates,
+        )
+
+    expected = {
```
ajrcampbell: The regression test exercises only fully hydrogenated, internal `LYS` and `ARG`, while the production rule covers 20 residue names, terminal variants, partial hydrogens and altloc selection.

Each of these plausible mutations leaves the complete Meeko suite green: restricting the standard-name set to `LYS` and `ARG`, requiring identical input and template atom inventories, ignoring terminal candidates, and disabling template graphs for altloc input. Separate probes show that each mutation changes live behaviour.

Please add compact table-driven cases for all standard names plus partial-hydrogen, terminal and selected-altloc paths. Each case should assert an independently expected named bond or fallback outcome.

meeko/utils/rdkitutils.py
ajrcampbell: Commit `f2eacf9c083de58dfdfa022adba008300cb9c647` now carries the validated PDB atom-name mapping into template matching and sorts consensus bonds before insertion. All four fixtures preserve exact named coordinates, and seeds 1 through 5 produce identical prepared PDB and PDBQT bytes. The full suite passes with 132 tests, four skips and two pre-existing warnings.

meeko/polymer.py
ajrcampbell: Commit `f1da3cc` returns a residue-local parse failure as unparsed so `allow_bad_res` owns the policy. A mixed valid and malformed receptor test retains `A:1` and identifies `Z:999`; the complete stack at `f2eacf9c083de58dfdfa022adba008300cb9c647` passes 132 tests with four skips.

meeko/polymer.py
ajrcampbell: Commit `f71bad5` forwards `set_template` into residue parsing and gives a valid explicit selection precedence. Both the custom `LYS` ambiguity case and input named `LYN` now select `LYS` and preserve named coordinates. The complete stack at `f2eacf9c083de58dfdfa022adba008300cb9c647` passes 132 tests with four skips.

test/polymer_creation_test.py
ajrcampbell: Commit `264160d` adds independent heavy-bond expectations for all 20 standard names plus partial-hydrogen, terminal, selected-altloc and coordinate-fallback cases. The stack also tests explicit selection, residue isolation and byte identity across five hash seeds. The complete suite at `f2eacf9c083de58dfdfa022adba008300cb9c647` passes 132 tests with four skips.

diogomart:
Throwing errors for distorted geometries is on purpose and valuable, so users can fix the problems in the molecules. However, this PR enables a valuable workflow: feeding distorted geometries to molecular minimization or short simulations.

This PR may become an optional code path alongside the other _"I don't care just move along"_ options, such as `--ignore_bad_res`.

